PR #59, opened 2017-02-22
Add Member Model

joaofcosta:
@martinhoaragao This is not finished yet, I'm opening a PR because I need your opinion on some questions. If you can check the current attributes for the Member model, notice that `has_one :responsible_member, class_name: 'Member'` referes to the member which registered/added the member (just like we have in our spreadsheet). I've also added `belongs_to :member` in the User model so later we can use something like `current_user.member`.

After checking that out do you think there's still need for `account_number`, `student_id, `city` and `phone_number` in the users table?

I would also suggest moving `user_type` from User to Member and changing it to `member_type`. While we're at it I would also think that it would be nice to have `admin` on the Member side, considering a user needs to be a member in order to be and admin.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -1,3 +1,12 @@
-class Member < User
-  has_one :payment, foreign_key: :user_id, dependent: :destroy
+class Member < ActiveRecord::Base
+  has_one :responsible_member, class_name: 'Member'
+  has_one :user
+  has_many :terms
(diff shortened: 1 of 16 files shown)


joaofcosta:
@martinhoaragao Can you check this now? :)

joaofcosta:
@martinhoaragao What about now? 🤔

PR #64, opened 2017-03-03
Create Department seeds.

joaofcosta:
@martinhoaragao Can you check if these are the names you want? I'm also wondering if you wish to create a `acronym` attribute that would be useful for departments like DREP and CAOS 🤔


joaofcosta:
@frmendes If my memory does not fail me we agreed that the departments shouldn't be dependent on the year or a board, because the departments do not change that much from year to year. A department will still be associated with a board through the terms that make up a board. 
This way we can avoid having to deal with boards and terms, which will let us have a usable website much faster.

PR #65, opened 2017-03-04
Remove UsersController inheriting from Clearance::UsersController.

joaofcosta:
* Clerance was already remove but this was left behind. Somehow, when
running on a local machine no error was shown, but when runninng on the
remote server the app would not run due to this.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,4 +1,4 @@
-class UsersController < Clearance::UsersController
+class UsersController < ApplicationController
   def new
     @user = User.new
   end


app/controllers/users_controller.rb
```diff
@@ -1,4 +1,4 @@
-class UsersController < Clearance::UsersController
+class UsersController
```
frmendes: It should inherit from `ApplicationController` instead.

joaofcosta: @frmendes Sorry, forgot to check the other controllers to check that. Thanks, fixed it 👍

PR #67, opened 2017-03-06
Add activities registration and creation with cancan.

joaofcosta:


config/routes.rb
```diff
@@ -1,7 +1,9 @@
 Rails.application.routes.draw do
   root 'welcome#index'
 
-  # resources :activities
+  resources :activities
+  get 'activities/:id/register', to: 'activities#register', as: 'activity_registration'
```
joaofcosta: @frmendes Should this be a post instead? 🤔

app/controllers/activities_controller.rb
```diff
@@ -35,6 +36,15 @@ def destroy
     redirect_to @activity
   end
 
+  def register
```
frmendes: What do you think of moving this to a different controller? Highly recommend reading [this](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/).

app/controllers/activities_controller.rb
```diff
@@ -35,6 +36,15 @@ def destroy
...
+    user = current_user
```
frmendes: Why do you need this assignment?

app/controllers/activities_controller.rb
```diff
@@ -35,6 +36,15 @@ def destroy
...
+    if !@activity.registrations.find_by(user_id: user.id)
```
frmendes: This should be something like `@activity.registered? current_user`

app/models/ability.rb
```diff
@@ -2,6 +2,12 @@ class Ability
   include CanCan::Ability
```
frmendes: Don't you need one for registrations?

config/routes.rb
frmendes: Depends on what you want to achieve. Both can work. Can you elaborate the use case?

joaofcosta: @frmendes I was thinking about, later in the process, use an AJAX request to register the user in an activity so we do not need to reload the entire page but only update the look of the button. What do you think about that?

app/controllers/activities_controller.rb
joaofcosta: @frmendes First of all, thanks for the link 👍 If I understand correctly you suggest creating a `Activity::RegistrationsController` with, possibly, only the `create` method? 🤔

app/controllers/activities_controller.rb
joaofcosta: @frmendes Agree with you on this one, much simpler :)

app/controllers/activities_controller.rb
joaofcosta: @frmendes I don't , it was just to avoid typing `current_user`, not my greatest idea now that I look at it again --.

app/models/ability.rb
joaofcosta: @frmendes I'm still trying to figure that out, I am trying to avoid using `except` in `load_and_authorize_resource` but I still haven't figured out the current permissions for which user so I can achieve that, will look into it a bit more and will get back to you ASAP.

app/controllers/activities_controller.rb
frmendes: Precisely

app/controllers/activities_controller.rb
frmendes: I'd recommend removing that assignment.

config/routes.rb
frmendes: Well, for now I think it's an overkill. Maybe in the future if it is justified.

I'd have a `get` for the signup form and a corresponding `post` for the `create` action.

joaofcosta: @frmendes There is no signup form right now, there's only a sign up button that links to this path and the redirects the user to the activity page.

app/controllers/activities_controller.rb
joaofcosta: @frmendes :+1:

app/controllers/activities/registrations_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Activities::RegistrationsController < ApplicationController
+  before_action :set_activity, only: [:create]
+
+  def create
+    if !@activity.registered?(current_user)
```
frmendes: Prefer `unless`

joaofcosta: @frmendes :+1:

joaofcosta:
@frmendes I've added permissions for Registration and refactored that little `if !` to `unless` :+1:

PR #70, opened 2017-03-31
Replace thin/webrick with puma

### LINKED ISSUE #26: Single threaded server in production
We are currently running a single threaded server in our production environment.

I would recommend changing it to puma and I wouldn't mind setting everything up.

frmendes:
Fixes #26


frmendes:
I added the gem but you'll have to double check it yourself ¯\_(ツ)_/¯

Issue #48, opened 2016-11-25
Responsive Images


joaofcosta:
Add responsive images so as to serve different image versions for different screen sizes.
Maybe using this gem https://github.com/johnkoht/responsive-images , haven't checked it out yet.

joaofcosta:
@fribmendes Closing this due to #71 👍

PR #71, opened 2017-03-31
Adds retina_tag

fribmendes:
This is an approach to fixing #48. [`retina_tag`](https://github.com/davydotcom/retina_tag) serves two versions of the image (img.png and img@2x.png, so it's up to you to add both) depending on the screen dpi.

To demonstrate I copied `uminhocup.jpg` to `uminhocup@2x.jpg` and reduced the image size of the first. One is served correctly on Retina screens, the other on regular screens.


frmendes:
@joaofcosta I thought so as well, but I had the impression I'd already used it without the `height` attribute. I tested and it used the '@2x' version on my end.

Here's what you can do. Checkout this branch, open the website on your computer and check if the UMinho Cup image is the '@2x' version. Then move it to one of those screens on CeSIUM and refresh. The same image should not be the '@2x' version. Can you double check it and let me know, please?

PR #74, opened 2017-04-13
Add a goof to the team's page 😎

joaofcosta:


joaofcosta:
@fribmendes The image? Yes. That's a recurring question in every PR which has new images, I think I'm gonna create a label specifically to tell if the images are already optimised or not 😆

joaofcosta:
@fribmendes ping 👍

PR #76, opened 2017-04-17
Move .env to capistrano's shared and update .env.example

martinhoaragao:
Move envs to shared file through Capistrano

Previously the .env file would have to be restored after each deployment. 
Moving it to the capistrano’s shared folder simplifies the process.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,7 @@
 # In order for the webapp to work correctly create your own .env file and fill
 # the credentials that you can see below.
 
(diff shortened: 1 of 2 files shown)


joaofcosta:
@fribmendes Sorta of, somehow the commit that added the `.env` file to Capistrano's shared folder just disappeared, poof!

martinhoaragao:
@fribmendes Yes you are. Somehow a commit vanished, pls don't ask me how 😨 .
On a good side, at least we now have 2 separate PRs and the .dot.example is not forgotten!

PR #81, opened 2017-04-19
Add activity registration confirmation

joaofcosta:


joaofcosta:
@fribmendes I remember that blog post, I was just trying to rush through this to get a working version as soon as possible, will move those methods to `index` and `update` methods inside the registrations controller and update the respective routes.

joaofcosta:
@fribmendes Check it again if possible :)

app/models/activity.rb
```diff
@@ -32,4 +32,8 @@ def end_date_is_after_start_date
   def registered?(user)
     self.registrations.find_by(user_id: user.id)
   end
+
+  def participants
+    User.find(self.registrations.map { |r| r.user_id })
```
fribmendes: `find` will only return the first record. You want to use `where`.

```ruby
User.where(id: registrations.map(&:user_id))
```

Or even better

```ruby
registrations.flat_map(&:user)
```

Wouldn't this work as well?

joaofcosta: @fribmendes `User.find` can actually receive an array of ids and will return an array with all the users (which is what we want here). Using `User.where` only results in an error `undefined method include?`.
I'm still going to change .map to .flat_map because it's cleaner, nice tip :+1:

app/models/activity.rb
```diff
@@ -32,4 +32,8 @@ def end_date_is_after_start_date
   def registered?(user)
     self.registrations.find_by(user_id: user.id)
   end
+
+  def participants
+    User.find(self.registrations.flat_map(&:user_id))
```
fribmendes: Well, according to [this](https://github.com/cesium/atomic/blob/master/app/models/registration.rb#L3), an instance of `Registration` has a `user` method. So why not `registrations.flat_map(&:user)`? The way you are putting it, `flat_map` is redundant. `User.find(registrations.map(&:user_id)` yields the same results without a call to flattening it.

joaofcosta: @fribmendes Hum, yeah `registrations.flat_map(&:user)` works! I am not well versed in `flat_map`, will update soon :+1:

fribmendes: It's basically the same as `array.map.flatten` but in one pass. Slightly optimized.

joaofcosta: @fribmendes Updated it! Also took the time to remove `self` since it isn't needed.

PR #82, opened 2017-04-25
Unfuck brainfart

fribmendes:


fribmendes:
@joaofcosta can you double check this works? Don't have the test data for it.

joaofcosta:
@fribmendes I'll check it tomorrow morning with the test data I have. Seems like a really nice little brainfart #classy :+1:

app/models/activity.rb
```diff
@@ -1,6 +1,8 @@
 class Activity < ActiveRecord::Base
   has_many :registrations
-  has_many :users, through: :registrations
+  has_many :participants, through: :registrations,
+                          class_name: "User"
```
joaofcosta: @fribmendes Using `class_name: "User"` I get the following error: `Could not find the source association(s) "participant" or :participants in model Registration. Try 'has_many :participants, :through => :registrations, :source => <name>'. Is it one of activity or user?`. Changing it to `source: :user` fixes the problem :+1:

fribmendes: @joaofcosta done

PR #83, opened 2017-08-25
Fix database migrations

joaofcosta:
🔨


martinhoaragao:
As far as I know, there's no live data currently @fribmendes. And only me and @joaofcosta have the current state of migrations, perhaps you too.

Furthermore the migrations are broken, and new devs can't run them from scratch properly. I don't like the fix. But it works, and only fucks me and @joaofcosta over.

db/schema.rb
```diff
@@ -115,5 +119,7 @@
 
   add_index "users", ["email"], name: "index_users_on_email", using: :btree
```
fntneves: Is there any statement that forces this column to be unique?

Moreover, if you require uniqueness on this column, I believe ActiveRecord will add an index.

martinhoaragao: I don't get what do you mean. 
Any article/documentation you can point me towards?

fntneves: `Adding a unique constraint will automatically create a unique B-tree index on the column or group of columns listed in the constraint.` in https://www.postgresql.org/docs/9.1/static/ddl-constraints.html

fntneves: Sorry, I read now that this is a generated file. 👍

PR #85, opened 2017-08-30
Improve Activity Registrations

joaofcosta:
### Description

This Pull Request pretends to improve the flow for user registrations, as of right now a user can only register for an activity, but if they clicked on the button without meaning it they have no way of canceling their registration.

On the admin side when an admin confirms the presence of an user on an activity he has no way of reverting that confirmation, which is bad because it might have ben an accident.

This Pull Request tackles those two problems.

### Changes

* Add `confirmed` field to `Registration`.

* Update `activities/show` to show a 'Cancel Registration' button so users can cancel a registration to an activity.

  * Create `destroy` method in `Activities::RegistrationsController`.

* Update `activities/registrations/index` view to display a 'Cancel Registration' button so activitiy admins can cancel confirmations if they ever make a mistake or simply if they need to.

  * Refactor `update` method of `Activities::RegistrationControoler` to simply reverse the `confirmed` field, allowing this method to be used for confirmation and denial of a user presence in a activity.

### DIFF
--- a/app/controllers/activities/registrations_controller.rb
+++ b/app/controllers/activities/registrations_controller.rb
@@ -20,6 +20,13 @@ class Activities::RegistrationsController < ApplicationController
       .find_by(user_id: registration_params[:participant_id])
     registration.update_attribute(:confirmed, !registration.confirmed)
 
+    if registration.confirmed
+      flash[:success] =
+        "Confirmed #{User.find(registration_params[:participant_id]).name}!"
+    else
+      flash[:alert] =
+        "Canceled #{User.find(registration_params[:participant_id]).name} confirmation!"
+    end
     redirect_to activity_participants_path
   end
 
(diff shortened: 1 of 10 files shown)


app/controllers/activities/registrations_controller.rb
```diff
@@ -15,13 +15,21 @@ def index
   end
 
   def update
-    @activity
+    registration = @activity
       .registrations
       .find_by(user_id: registration_params[:participant_id])
-      .update_attributes(confirmed: true)
+    registration.update_attribute(:confirmed, !registration.confirmed)
```
fribmendes: This is dangerous though. If I refresh while sending the request or click the button twice (normal in slow internet like the university), what will happen? I might think I signed up but instead signed up and then signed out.

The button should post a `confirmation` parameter

marcelo140:
Fixed in https://github.com/cesium/atomic/pull/95

PR #87, opened 2017-09-13
Enables login

marcelo140:
This PR adds the following changes:
- Adds login/logout button to the page
- Removes facebook authentication as requested
- Redirects users to the welcome page if they try to sign in while authenticated

**Question:** To remove the facebook authentication I guided myself by https://github.com/cesium/atomic/pull/52/ where there was a discussion about the `auth_email` in the user model. It was stated that the method was needed since the user could choose not to share his email when logging in with facebook. Is this method needed for github/email authentication which were added later? Can the users opt to not share the email with these two providers? 

**Note:** This PR should not be merged to master until we confirm that the activities are working as intended.


martinhoaragao:
@joaofcosta could you take a look at the auth_email that he mentioned in the PR description? 
I think it can be removed, as I can only recall Facebook allowing one to not provide the email. Everything other than that seems rather fine.

app/controllers/sessions_controller.rb
```diff
@@ -1,4 +1,8 @@
 class SessionsController < ApplicationController
+  def new
+    redirect_to root_url unless session[:user_id].nil?
```
fribmendes: I'd rather have this done with `pandit` or `cancancan`, tbh

marcelo140: The project uses `cancancan` but wouldn't I need to have a `session` model to use it here?

fribmendes: Two options:

1. which I'm not really a fan of:

```ruby
before_filter -> { raise CanCan::AccessDenied if current_user }
```

2. What's described in here: https://github.com/ryanb/cancan/issues/22

marcelo140:
Regarding to the `auth_email` method, I checked our current database and @9thScientist, who logged in with github, has an email provided by this method which is even weirder since he has a public email on github.

fribmendes:
@marcelo140 not on par with what is happening but don't forget GitHub allows private emails as well.

PR #88, opened 2017-09-18
Add activities to factory girl

marcelo140:


spec/factories/activity.rb
```diff
@@ -0,0 +1,11 @@
+FactoryGirl.define do
+  factory :activity do
+    name { Faker::Hacker.abbreviation }
```
martinhoaragao: Can you use longer titles please? One of the useful instances of factory generated content is ensuring we have non-fixed length data that simulates production. 

Activity names can be rather large. Such as one in a few weeks "Ask Hell. What do you mean?", how would this fare in a mobile responsive web view of that page? Abbreviations are too short to achieve this.

> Faker::BossaNova.song

Might be enough.

PR #89, opened 2017-09-18
Adds create activity button

9thScientist:
This PR adds the following changes:
- Adds a "Atividades" option on the navbar
- Adds a "Criar Atividade" button that only shows to users who have the ability to create activities. 
- Fixes minor bugs

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -1,5 +1,4 @@
 class ActivitiesController < ApplicationController
-  before_action :set_activity, except: [:index]
   load_and_authorize_resource
 
   def new
@@ -12,6 +11,7 @@ class ActivitiesController < ApplicationController
   end
 
   def show
+    @activity = Activity.find(params[:id])
   end
 
   def create
@@ -24,6 +24,8 @@ class ActivitiesController < ApplicationController
   end
 
   def update
+    @activity = Activity.find(params[:id])
+
     if @activity.update(activity_params)
       redirect_to @activity
     else
@@ -32,17 +34,16 @@ class ActivitiesController < ApplicationController
   end
 
   def destroy
-    @activity.destroy
-    redirect_to @activity
+    Activity.find(params[:id]).try(:destroy)
+    redirect_to activities_path
   end
 
   private
-    def activity_params
-      params.require(:activity).permit(:name, :location, :description, :total_rating, :member_cost, :guest_cost, :start_date, :end_date, :coffee_break, :poster, :department_id)
-    end
 
-    def set_activity
-      @activity = Activity.find(params[:id])
-    end
+  def activity_params
+    params.require(:activity).permit(:name, :location, :description,
+      :total_rating, :member_cost, :guest_cost, :start_date, :end_date,
+      :coffee_break, :poster, :department_id)
+  end
 
 end
--- a/app/views/activities/_form.slim
+++ b/app/views/activities/_form.slim
@@ -1,4 +1,11 @@
 = form_for @activity, html: {multipart: true} do |form|
+
+  - if @activity.errors.any?
+    div.alert class="alert-danger"
+      - error_messages(@activity).each do |msg|
+        li
+          = msg
+
   = form.label :name,
     class: "activity-form-label"
   br
@@ -27,12 +34,14 @@
     class: "activity-form-label"
   br
   = form.number_field :member_cost,
+    value: 0,
     class: "activity-form-field"
 
   = form.label :guest_cost,
     class: "activity-form-label"
   br
   = form.number_field :guest_cost,
+    value: 0,
     class: "activity-form-field"
 
   = form.label :start_date,
--- a/app/views/activities/index.slim
+++ b/app/views/activities/index.slim
@@ -1,6 +1,12 @@
 = render partial: 'activities/activities_breadcrumb'
 .divide60
 .container.blog-left-img
+
+  - if can_create?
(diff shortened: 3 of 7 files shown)


app/controllers/activities_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ActivitiesController < ApplicationController
-  before_action :set_activity, except: [:index]
+  before_action :set_activity, except: [:index, :new, :create]
```
marcelo140: I think it would be more readable with `only: [:show, :destroy]` or, even better, defining the variable in the methods it's needed.

app/views/activities/show.slim
```diff
@@ -14,7 +14,7 @@
         = simple_format(@activity.description, class: "lead")
         = render partial: "shared/activity/panel", object: @activity
         = link_to "Inscrever-me!",
-          (@activity.registration_link != nil) ? @activity.registration_link : "#",
+          ( activity_registration_path(@activity) != nil) ? activity_registration_path(@activity) : "#",
```
marcelo140: Just want to point out that ruby has a method `nil?`. Not that important though.

app/controllers/activities_controller.rb
```diff
@@ -24,6 +24,7 @@ def create
   end
 
   def update
+    @activity = Activity.find(params[:id])
     if @activity.update(activity_params)
```
fribmendes: Newline missing

app/controllers/activities_controller.rb
```diff
@@ -32,6 +33,7 @@ def update
   end
 
   def destroy
+    @activity = Activity.find(params[:id])
     @activity.destroy
     redirect_to @activity
```
fribmendes: If you are destroying it you won't be able to redirect to it.

app/views/activities/index.slim
```diff
@@ -1,6 +1,12 @@
 = render partial: 'activities/activities_breadcrumb'
 .divide60
 .container.blog-left-img
+
+  - if can? :create, Activity
```
fribmendes: This should be a helper method

app/views/activities/show.slim
```diff
@@ -32,7 +32,7 @@
                 activity_registration_path(@activity.id),
                 class: 'btn border-theme btn-lg'
         hr
-        - if current_user.activity_admin?
+        - if current_user != nil && current_user.activity_admin?
```
fribmendes: `!= nil` is redundant

app/controllers/activities_controller.rb
9thScientist: I also thought that it was strange. Do you want me to change it so that it redirects to index?

app/controllers/activities_controller.rb
9thScientist: :+1:

app/controllers/activities_controller.rb
fribmendes: Yes please

app/controllers/activities_controller.rb
```diff
@@ -32,17 +34,15 @@ def update
   end
 
   def destroy
+    @activity = Activity.find(params[:id])
+
     @activity.destroy
```
fribmendes: You don't need to instantiate. I would change this to `Activity.find(params[:id]).try(:destroy)`

app/controllers/activities_controller.rb
```diff
@@ -32,17 +34,15 @@ def update
...
-    redirect_to @activity
+    redirect_to activities_path
   end
 
   private
     def activity_params
       params.require(:activity).permit(:name, :location, :description, :total_rating, :member_cost, :guest_cost, :start_date, :end_date, :coffee_break, :poster, :department_id)
```
fribmendes: Since you are working with in this file, can you break a line somewhere please?

app/controllers/activities_controller.rb
```diff
@@ -32,17 +34,15 @@ def update
...
   def destroy
+    @activity = Activity.find(params[:id])
+
     @activity.destroy
-    redirect_to @activity
+    redirect_to activities_path
   end
 
   private
     def activity_params
```
fribmendes: This shouldn't be idented this far.

app/views/activities/_form.slim
```diff
@@ -1,4 +1,11 @@
 = form_for @activity, html: {multipart: true} do |form|
+
+  - if @activity.errors.any?
+    div.alert class="alert-danger"
+      - @activity.errors.full_messages.each do |msg|
```
fribmendes: This could be in a helper method

app/views/activities/show.slim
fribmendes: You could do `activity_registration_path(@activity) || "#"`

app/controllers/activities_controller.rb
9thScientist: What do you mean? The indentation seems right.

app/controllers/activities_controller.rb
```diff
@@ -34,15 +34,16 @@ def update
   end
 
   def destroy
-    @activity = Activity.find(params[:id])
+    @activity = Activity.find(params[:id]).try(:destroy)
```
fribmendes: you don't need to instantiate, the variable is not used.

app/controllers/activities_controller.rb
fribmendes:
```ruby
private

def stuff
  # ...
end
```

The indentation is in the same column

PR #91, opened 2017-09-20
Split activities according to it's date

marcelo140:


app/views/activities/index.slim
```diff
@@ -1,9 +1,17 @@
...
+      a href==activities_path(show: "next")
+        | Próximas atividades
+    li
+      a href==activities_path(show: "previous")
+        | Atividades realizadas
+
 .divide60
 .container.blog-left-img
-  - if @activities.empty?
-    h4 Atividades não encontradas
```
9thScientist: Is there a good reason why you are removing this message? It could be useful when there's no next activities.

app/views/activities/index.slim
```diff
@@ -1,9 +1,17 @@
...
   - @activities.each do |activity|
     = render partial: "shared/activity/index", locals: { activity: activity }
 
-  = will_paginate @activities, renderer: BootstrapPagination::Rails
+  = will_paginate @next_events, renderer: BootstrapPagination::Rails
```
9thScientist: Where is `@next_events` declared?

app/controllers/activities_controller.rb
```diff
@@ -7,8 +7,17 @@ def new
   end
 
   def index
-    @activities = Activity.all.order('start_date DESC')
-      .paginate(page: params[:page], per_page: 10)
+    if (params[:show].nil? || params[:show] == "next")
+      @activities = Activity
```
fribmendes: Please extract this to a scope in `activity.ex`

app/views/activities/index.slim
```diff
@@ -1,9 +1,17 @@
 = render partial: 'activities/activities_breadcrumb'
+
+.container.navbar.navbar-default
+  ul.nav.navbar-nav
+    li
+      a href==activities_path(show: "next")
+        | Próximas atividades
+    li
+      a href==activities_path(show: "previous")
```
fribmendes: The `==` works?

marcelo140: Yes it does. It's described in the [slim documentation](http://www.rubydoc.info/gems/slim/frames#Ruby_attributes).

fribmendes: TIL

app/models/activity.rb
```diff
@@ -19,13 +19,17 @@ class Activity < ActiveRecord::Base
...
-    order('start_date DESC').
-    limit(5).
-    where("activities.end_date >= ?", Time.now)
+    where('? < activities.end_date', DateTime.current).
+    order('start_date ASC')
+  }
+
+  scope :previous_activities, -> {
+    where('? > activities.end_date', DateTime.current).
+    order('start_date DESC')
```
fribmendes: Looks great! Now add tests for this please

app/views/activities/index.slim
marcelo140: I may change this to a `link_to` though

fribmendes: Would be more sensible, yes

app/controllers/activities_controller.rb
```diff
@@ -7,8 +7,13 @@ def new
   end
 
   def index
-    @activities = Activity.all.order('start_date DESC')
-      .paginate(page: params[:page], per_page: 10)
+    if (params[:show].nil? || params[:show] == "next")
```
fribmendes: Reviewed this again, I think this should be extracted to a method, since it's more self describing. I don't know what it is supposed to be doing just by looking at the conditions, tbh

spec/factories/activity.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :activity do
-    name { Faker::BossaNova.song }
+    name { Faker::App.name }
```
marcelo140: Apparently BossaNova isn't a symbol in Faker.... my bad

martinhoaragao: It is, you must have the gem outdated though.
Could you update Faker? 😂

marcelo140: Oops 😲 
Creating a new PR to update it ....

app/controllers/activities_controller.rb
```diff
@@ -45,4 +50,7 @@ def set_activity
       @activity = Activity.find(params[:id])
     end
 
+    def previous_activities_requested?
+      params[:show] == "previous"
```
fribmendes: me likes

spec/models/activity_spec.rb
```diff
@@ -40,35 +42,35 @@
...
+    it "end in the future" do
+      activities = Activity.next_activities
+      expect(activities.none?(&:already_happened?)).to be true
     end
 
-    describe "when name is too long" do
-      before { @activity.name = 'a' * 76 }
-      it { should_not be_valid }
-    end
+    it "are sorted by nearest date first" do
```
fribmendes: "it are sorted" doesn't read well.

PR #95, opened 2017-09-23
Improve Activity Registrations 

marcelo140:
I create a branch from https://github.com/cesium/atomic/pull/85 to fix the last issues. My changes on @joaofcosta branch are:

- change the registration#update action to make it idempotent
- fix an issue where users could cancel their registration
- extract the logic from a view to a partial

**In case you review changes per commit:** the registration controller changes where supposed to be added in commit deb562d and not in 181f43d.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -3,12 +3,15 @@ class Ability
 
   def initialize(user)
     user ||= User.new # guest user (not logged in)
-    if user.activity_admin?
-      can :manage, [Activity, Registration]
-    else
+
+    case user.permissions
+    when :guest
       can :read, Activity
-      can [:read, :create], Registration
+    when :user
+      can :read, Activity
+      can [:create, :destroy], Registration, user_id: user.id
+    when :activity_admin
+      can :manage, [Activity, Registration]
     end
   end
-
 end
--- a/app/views/activities/_registration.slim
+++ b/app/views/activities/_registration.slim
@@ -1,0 +1,10 @@
+- if @activity.end_date.future?
+  - if can?(:destroy, Registration) && @activity.registered?(current_user)
+    = link_to activity_registration_cancel_path(@activity.id),
+      class: "btn btn-disabled btn-lg" do
+      i.fa.fa-times.fa-fw
+      | Anular Inscrição
+  - elsif can?(:create, Registration)
+    = link_to "Inscrever-me!",
+      activity_registration_path(@activity) || "#",
+      class: "btn border-theme btn-lg" + (@activity.end_date.past? ? " disabled": "")
--- a/app/views/activities/registrations/index.slim
+++ b/app/views/activities/registrations/index.slim
@@ -1,19 +1,32 @@
-= render partial: 'activities/activities_breadcrumb'
+= render partial: "activities/activities_breadcrumb"
 .divide60
 .container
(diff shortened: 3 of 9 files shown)


app/views/activities/registrations/index.slim
```diff
@@ -1,19 +1,32 @@
...
-      .col-md-2.col-sm-4.col-xs-12
-        - if @registrations[index].confirmed
-          button.btn.btn-disabled.pull-right Confirmado
-        - else
-          = link_to 'Confirmar',
-            confirm_participant_path(@activity.id, participant.id),
-            class: 'btn border-theme pull-right'
+      .col-md-4.col-sm-4.col-xs-12
+        = form_for @registrations[index], 
+          url: confirm_participant_path(@activity.id, participant.id) do |f|
```
marcelo140: I actually have a question about this route. Is the way we pass the `acitivity_id` and the `participant_id` instead of the `registration_id` some kind of convention for url readiblity?

marcelo140:
@joaofcosta came back to life, I don't think we need this.

app/views/activities/_registration.slim
```diff
@@ -0,0 +1,10 @@
+- if current_user.present? && @activity.end_date.future?
```
fribmendes: Do you need the `current_user` verification here? Shouldn't this be handled by cancancan or pandit or wtv?

app/views/activities/registrations/index.slim
fribmendes: Is it possible that you might not have a registration at this point?

marcelo140: The point of this form is to confirm the presence of a registered user by updating the registration therefore the registration must already exist.

fribmendes: What happens when you use the id?

marcelo140: I don't know if the question has a trick but when I use the id it just works. Btw, the `participant` is a relation that exists through the registration model so the registration must must exist either way.

marcelo140:
shameless ping at a sunday @fribmendes

app/models/ability.rb
```diff
@@ -3,12 +3,15 @@ class Ability
...
-      can :manage, [Activity, Registration]
-    else
+
+    case user.permissions
+    when :guest
+      can :read, Activity
+    when :user
       can :read, Activity
-      can [:read, :create], Registration
+      can [:create, :destroy], Registration
```
fribmendes: Update, no?

marcelo140: Nope, an user can only register or cancel it's registration it's own registration. Only the admin can update (confirm) his presence in the activity.

I think I forgot to add an hash condition so that an user can only create/destroy his own registrations though.

marcelo140:
Fixed several bugs found in staging (huh, maybe we should have tested before requesting a review):

- Two successive unregistrations would try to delete the resource twice, causing an user error.
- Confirmation buttons showed in front of the wrong participant due to the fact that two different arrays were being used.
- The order of registrations in the index would changed with confirmations/cancellations.
- Apparently I was wrong in previous discussion. When an admin tries to confirm the participation of an user, the registration may no longer exist if the user cancelled it by himself.

@fribmendes tell me something if you wish to review again, otherwise I will merge it.

PR #97, opened 2017-10-11
Update welcome

marcelo140:
Didn't delete the "Welcome" image because I'm expecting to change back to it once codeweek is over. Am I wrong?


marcelo140:
Forgot master has changes that should not be in live. Closing this.

PR #98, opened 2017-10-11
Update welcome

marcelo140:
Didn't delete the "Welcome" image because I'm expecting to change back to it once codeweek is over. Am I wrong?


app/views/welcome/index.html.slim
```diff
@@ -3,9 +3,11 @@
   .tp-banner
     ul
       /! SLIDE
-      li data-masterspeed="1000" data-slotamount="5" data-title=("Bem-vindo") data-transition="fade"
+      li data-masterspeed="1000" data-slotamount="5" data-title=("Codeweek") data-transition="fade"
```
martinhoaragao: Can you change it to "CodeWeek"?

app/views/welcome/index.html.slim
```diff
@@ -3,9 +3,11 @@
...
         /! MAIN IMAGE
-        = carousel_image_tag('boasvindas_v9.png')
+        = carousel_image_tag('codeweek.png')
+        = slider_link(start_second: 400, x: 1000, y: 480, title: "Ver mais",
```
martinhoaragao: It needs a target _blank to open in a new tab instead. Not sure how it's done with the slider_link, I thought it had already been used as such.

marcelo140: I was the one who created this helper. Fixed it already.

martinhoaragao: Cool. Got it confused with another helper then. Go ahead!

PR #99, opened 2017-10-12
Adds default poster to activities index and show

9thScientist:
This PR adds a default poster to the list of activities and the activity itself. Before, if a image had no poster, it would look completely different from the other activities, which made the whole index inconsistent.


app/views/activities/show.slim
```diff
@@ -6,20 +6,16 @@
...
-        = render partial: "shared/activity/panel", object: @activity
-        = render partial: "registration", object: @activity
-    - else
-      .col-md-6
-        = simple_format(@activity.description, class: "lead")
-      .col-md-6
-        = render partial: "shared/activity/panel", object: @activity
-        = render partial: "registration", object: @activity
+      - else
+        = image_tag "poster_default.png", class: "img-responsive"
```
marcelo140: This `if poster exists then { image  } else { default image }` is used twice. You could abstract this to a helper similar to `activity_poster(activity, class)`.

app/views/shared/activity/_index.slim
```diff
@@ -1,16 +1,11 @@
...
-      = render partial: "shared/activity/description_date", locals: { activity: activity }
-- else
-  h3.activity-title
-    = link_to activity.name, activity_path(activity)
-  .row
-    .col-md-12
-      = render partial: "shared/activity/description_date", locals: { activity: activity }
-br
+    - else
+      = link_to image_tag("poster_default.png", class: "img-responsive activity-poster"), activity
```
marcelo140: This `if poster exists then { image  } else { default image }` is used twice. You could abstract this to a helper similar to `activity_poster(activity, class)`.

marcelo140:
A more simplistic approach would be to simply set the `default_poster` as the default value of `poster`.

PR #100, opened 2017-10-12
Adds meta tags to activities

9thScientist:
This PR adds meta tags to each activity, this way the activity's link can be shared on other websites so that they preview the activity's poster, title and description. The description is limited to 140 characters.

### DIFF
--- a/app/views/activities/show.slim
+++ b/app/views/activities/show.slim
@@ -1,4 +1,10 @@
 = render partial: 'activities/activities_breadcrumb'
+
+- content_for :head do
+  meta property='og:image' content='#{@activity.poster.url}'
+  meta property='og:title' content='#{@activity.name}'
+  meta property='og:description' content='#{truncate(@activity.description, length: 140)}'
+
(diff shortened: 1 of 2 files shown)


app/views/activities/show.slim
```diff
@@ -1,4 +1,11 @@
 = render partial: 'activities/activities_breadcrumb'
+
+- content_for :head do
+  - if @activity.poster.exists?
+    meta property='og:image' content='#{@activity.poster.url}'
```
marcelo140: Btw @martinhoaragao, are you okay with not to showing the default poster in previews?

martinhoaragao: Oh, you're right. We do want to have an image when we share the link, so there's no reason to hide the image although it's just a placeholder.

9thScientist: As soon as #99 is merged, `activity.poster.url` will show the default picture. That said, I have to remove that `if` in order to actually show the default poster.

PR #101, opened 2017-10-13
Add more registration options

marcelo140:
This PR adds two new columns to an activity:

- **external_link**: allows the admins to specify an external registration platform in case they need to use a more complex process (ex: doodle)

- **allows_registrations**: the admins can now specify if the users can start signing up for the activity or not. This allows the admins to announce an event earlier even if they still don't know exactly when will it happen.

### DIFF
--- a/app/views/activities/_registration.slim
+++ b/app/views/activities/_registration.slim
@@ -1,10 +1,15 @@
-- if @activity.end_date.future?
-  - if can?(:destroy, Registration) && @activity.registered?(current_user)
+- if @activity.external_link.empty?
+  - if can_cancel_registratation?(@activity)
     = link_to activity_registration_cancel_path(@activity.id),
       class: "btn btn-disabled btn-lg" do
       i.fa.fa-times.fa-fw
       | Anular Inscrição
-  - elsif can?(:create, Registration)
+  - elsif can_registrate?(@activity)
     = link_to "Inscrever-me!",
       activity_registration_path(@activity) || "#",
-      class: "btn border-theme btn-lg" + (@activity.end_date.past? ? " disabled": "")
+      class: "btn border-theme btn-lg"
+- elsif can_registrate?(@activity)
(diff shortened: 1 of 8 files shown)


app/views/activities/_registration.slim
```diff
@@ -1,10 +1,17 @@
-- if @activity.end_date.future?
-  - if can?(:destroy, Registration) && @activity.registered?(current_user)
+- registration = @activity.registrations.build
+
+- if @activity.external_link.empty?
+  - if can?(:destroy, registration) && @activity.registered?(current_user)
```
9thScientist: This condition should probably be a helper method.

app/views/activities/_registration.slim
```diff
@@ -1,10 +1,17 @@
...
       class: "btn btn-disabled btn-lg" do
       i.fa.fa-times.fa-fw
       | Anular Inscrição
-  - elsif can?(:create, Registration)
+  - elsif can?(:create, registration)
     = link_to "Inscrever-me!",
       activity_registration_path(@activity) || "#",
-      class: "btn border-theme btn-lg" + (@activity.end_date.past? ? " disabled": "")
+      class: "btn border-theme btn-lg"
+- elsif can?(:create, registration)
```
9thScientist: @fribmendes will probably want this in a helper method as well.

db/schema.rb
```diff
@@ -31,9 +31,11 @@
     t.integer  "poster_file_size"
     t.datetime "poster_updated_at"
     t.integer  "activity_id"
-    t.datetime "created_at",                                             null: false
-    t.datetime "updated_at",                                             null: false
+    t.datetime "created_at",                                                             null: false
+    t.datetime "updated_at",                                                             null: false
```
9thScientist: Nothing wrong here, but why did you change the indentation?

marcelo140: It was generated

PR #102, opened 2017-10-13
Adds news link to navbar

9thScientist:
This PR just adds a link to `cesium.di.uminho.pt/noticias` to the navbar.

### DIFF
--- a/app/views/layouts/lp/_navbar.html.slim
+++ b/app/views/layouts/lp/_navbar.html.slim
@@ -14,8 +14,8 @@
           = link_to 'Sobre', about_path
         li
           = link_to 'Equipa', team_path
-        / li
-        /   = link_to 'NOTÍCIAS', news_path
+        li
+          = link_to 'NOTÍCIAS', 'http://cesium.di.uminho.pt/news'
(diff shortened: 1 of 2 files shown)


marcelo140:
All the routes are in english so I think it would be a better idea to ask jmv to change `/noticias` to `/news` and disable the atomic `/news` route that is not fully developed yet (*I think*). This way it would still be the same route once we move the news to atomic.

By the way, the `http://cesium.di.uminho.pt/news` is more developed than I expected. Do you have any idea what needs to be done in order to be usable @martinhoaragao ?

martinhoaragao:
Absolutely agree with @marcelo140 .
Regarding the status of /news we should meet up to discuss it.

martinhoaragao:
What's delaying this?

9thScientist:
@marcelo140 @martinhoaragao Is this PR ready to me merged?

marcelo140:
As I've already told you I don't know if the `link_to` works and since you didn't test it I can't give you my approval.

marcelo140:
Woops, my bad.

PR #117, opened 2017-10-27
Change prh's title from "diretor de curso" to "diretor de departamento de informática"

machadovilaca:


machadovilaca:
to do that, wouldn't be better to use "Diretor de Curso ('Start' - 'End')"?

marcelo140:
Apparently "Department Director" is more important than "Course Director" so our "very respectable" president wants us to present the highest role.

machadovilaca:
https://screenshots.firefox.com/vkwio0ienPFp8jr9/cesium.di.uminho.pt
https://screenshots.firefox.com/MXB1BdoAgZZ63eVN/cesium.di.uminho.pt

all on the same line or the name and the title separated?

marcelo140:
Second option.

machadovilaca:
@marcelo140

app/assets/stylesheets/landing_page/_base.scss
```diff
@@ -171,6 +171,11 @@ p.lead{
     font-size: 16px;
     font-weight: 400;
 }
+
+p.bottom-space{
```
martinhoaragao: This seems unneeded. A new class just to remove the bottom margin of a paragraph?

app/views/welcome/about.html.slim
```diff
@@ -25,8 +25,9 @@
         |  (...) A partir desse momento iniciou-se o caminho deste núcleo estudantil que, 20 anos
         |  depois, mantém o nome e sigla e mantém o propósito determinado na projeção da imagem
         |  do curso e na defesa dos interesses dos seus associados."
-      p.lead.text-right
-        | - Pedro Rangel Henriques, Diretor de Curso de MiEI
+      p.lead.text-right.bottom-space Pedro Rangel Henriques
+      p.lead.text-right Diretor do Departamento de Informática - 2017
```
martinhoaragao: Is there a reason for a new paragraph here? 

Perhaps you could use a break and keep it as a single paragraph. Ideally, the styling should be different between the name and the title. But in this situation, for an easier fix, a single paragraph with a break seems to achieve quite the same style.

Look up here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

app/assets/stylesheets/landing_page/_base.scss
```diff
@@ -171,6 +171,11 @@ p.lead{
...
+    margin:2px;
```
martinhoaragao: Lacks a space.

app/views/welcome/about.html.slim
machadovilaca: Understood, it's a way better fix. 
I was thinking about the title as a new paragraph therefore I wasn't thinking about using the line break tag as it is not recommended.

app/views/welcome/about.html.slim
```diff
@@ -25,8 +25,8 @@
         |  (...) A partir desse momento iniciou-se o caminho deste núcleo estudantil que, 20 anos
         |  depois, mantém o nome e sigla e mantém o propósito determinado na projeção da imagem
         |  do curso e na defesa dos interesses dos seus associados."
-      p.lead.text-right
-        | - Pedro Rangel Henriques, Diretor de Curso de MiEI
+      p.lead.text-right Pedro Rangel Henriques</br>Diretor do Departamento de Informática - 2017
```
marcelo140: We don't mix html with slim. You can use the [verbatim operator](http://www.rubydoc.info/gems/slim/frames#Verbatim_text__) and intercalate it with a br tag:

```
p.lead.text-right
  | Prh
  br
  | Diretor
```
I think it will ok after this.

PR #120, opened 2017-10-30
Make login and activity registration more noticeable

9thScientist:
This PR aims to enhance UI/UX by making the login button more noticeable and draw attention to the fact that is possible to register in activities. Since you need to be logged in to show the 'register' button, people were not noticing it.

This PR should:
- [X] Make the 'login' button more noticeable
- [X] Make the 'register' button always present
- [ ] If the user isn't logged in, the 'register' button should redirect to the sign in page and, after a successful login, the user is redirected to the activity page.
- [ ] The 'logout' button should be replaced with the user image.
- [ ] The user image, once clicked, should present a dropdown menu with the option 'logout'

### DIFF
--- a/app/helpers/activities_helper.rb
+++ b/app/helpers/activities_helper.rb
@@ -19,6 +19,16 @@ module ActivitiesHelper
     can?(:create, registration) && !activity.registered?(current_user)
   end
 
+  def activity_link(activity)
+    if current_user.nil?
+      sign_in_path
+    elsif activity.external_link.empty?
+      activity_registration_path(@activity)
+    else
+      @activity.external_link
+    end
+  end
+
   def can_create?
     can? :create, Activity
   end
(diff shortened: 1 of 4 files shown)


app/views/activities/_registration.slim
```diff
@@ -8,8 +8,16 @@
...
+      class: "btn border-theme btn-lg"
 - elsif can_registrate?(@activity)
   = link_to "Inscrever-me!",
     @activity.external_link,
     class: "btn border-theme btn-lg",
     target: '_blank'
+- elsif current_user.nil?
+  = link_to "Inscrever-me!",
+    sign_in_path,
+    class: "btn border-theme btn-lg"
```
fribmendes: Duplicate code.

Also, consider extracting these to helpers, all those ifs in a view is not cool 😕

app/helpers/activities_helper.rb
```diff
@@ -9,6 +9,16 @@ def can_registrate?(activity)
...
 
+  def activity_link(activity)
+      if current_user.nil?
+	sign_in_path
+      elsif activity.external_link.empty?
+	activity_registration_path(@activity)
+      else
+	@activity.external_link
+      end
+  end
```
marcelo140: Indentation is wrong

PR #122, opened 2017-10-31
Add markdown support for activities description

ZePedroResende:
Added gem redcarpet for markdown support in activities description


app/helpers/activities_helper.rb
```diff
@@ -16,4 +16,13 @@ def can_create?
   def error_messages(activity)
     activity.errors.full_messages
   end
+
+  def markdown(text)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+        no_intra_emphasis: true,
+        fenced_code_blocks: true,
+        disable_indented_code_blocks: true)
```
marcelo140: Any reason to exclude the `autolink` option? It seems useful to me.

app/views/activities/show.slim
```diff
@@ -16,7 +16,7 @@
       = image_tag @activity.poster.url, class: "img-responsive"
       br
     .col-md-7
-      = simple_format(@activity.description, class: "lead")
+      = simple_format(markdown(@activity.description), class: "lead")
```
marcelo140: `simple_format` is creating an empty paragraph. The paragraph created by the markdown renderer is missing the `lead` class.

app/views/activities/show.slim
```diff
@@ -1,10 +1,9 @@
 = render partial: 'activities/activities_breadcrumb'
 
 - content_for :head do
-  meta property='og:image' content='#{@activity.poster.url}'
-  meta property='og:title' content='#{@activity.name}'
-  meta property='og:description' content='#{truncate(@activity.description, length: 140)}'
-
+  meta property='og:image' content='#{markdown(@activity.poster.url)}'
+  meta property='og:title' content='#{markdown(@activity.name)}'
+  meta property='og:description' content='#{markdown(@activity.description)}'
```
9thScientist: We used to truncate the `@activity.description` to a maximum of 140 characters, so that we wouldn't have all the description inside the meta tag. Furthermore, doesn't `markdown` helper just replace markdown syntax with html tags? Are you sure we can use those inside a meta tag? I think it would be more appropriate to use `plaintext` here.

app/views/activities/show.slim
```diff
@@ -1,10 +1,9 @@
 = render partial: 'activities/activities_breadcrumb'
 
 - content_for :head do
-  meta property='og:image' content='#{@activity.poster.url}'
-  meta property='og:title' content='#{@activity.name}'
-  meta property='og:description' content='#{truncate(@activity.description, length: 140)}'
-
+  meta property='og:image' content='#{markdown(@activity.poster.url)}'
```
9thScientist: This doesn't appear to make much sense. Are you trying to apply markdown to a image URL?

app/helpers/activities_helper.rb
```diff
@@ -16,4 +16,20 @@ def can_create?
   def error_messages(activity)
     activity.errors.full_messages
   end
+require 'redcarpet'
+require 'redcarpet/render_strip'
```
marcelo140: Ehh, any reason for this to be in the middle of the code?

app/helpers/activities_helper.rb
```diff
@@ -16,4 +16,20 @@ def can_create?
...
+
+  def markdown(text)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+        no_intra_emphasis: true,
+        fenced_code_blocks: true,
+        disable_indented_code_blocks: true,
+        autolink: true)
+    return markdown.render(text).html_safe
```
marcelo140: The explicit return is unnecessary.

app/helpers/activities_helper.rb
```diff
@@ -16,4 +16,20 @@ def can_create?
...
+  end
+
+  def plaintext(markdown)
+    plaintext= Redcarpet::Markdown.new(Redcarpet::Render::StripDown )
+    return plaintext.render(markdown)
```
marcelo140: The explicit return is unnecessary.

app/views/shared/activity/_description_date.slim
```diff
@@ -1,5 +1,5 @@
 p.lead.activity-description
-  | #{truncate(activity.description, length: 140)}
+  | #{truncate(plaintext(activity.description), length: 140)}
```
marcelo140: You can use a [ruby call](http://www.rubydoc.info/gems/slim/frames#Output__) to avoid the string interpolation.

Gemfile
```diff
@@ -18,6 +18,9 @@ gem 'animate-rails'
 gem 'wow-rails'
 gem 'owlcarousel-rails'
 
+# markdown
+gem 'redcarpet'
```
marcelo140: Well, I don't think we will have more markdown gems so could you move this to `atomic gems`.

app/helpers/activities_helper.rb
marcelo140: By the way, add an extra space before the `=` in line 32.

Gemfile
```diff
@@ -18,11 +18,13 @@ gem 'animate-rails'
 gem 'wow-rails'
 gem 'owlcarousel-rails'
 
+
```
fribmendes: ⚠️

app/helpers/activities_helper.rb
```diff
@@ -1,3 +1,6 @@
+require 'redcarpet'
+require 'redcarpet/render_strip'
```
fribmendes: You don't need to require IIRC.

app/helpers/activities_helper.rb
```diff
@@ -16,4 +19,18 @@ def can_create?
   def error_messages(activity)
     activity.errors.full_messages
   end
+
+  def markdown(text)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+        no_intra_emphasis: true,
```
fribmendes: Please use 2 space indentation.

app/helpers/activities_helper.rb
```diff
@@ -16,4 +19,18 @@ def can_create?
...
+        fenced_code_blocks: true,
+        disable_indented_code_blocks: true,
+        autolink: true)
+    markdown.render(text).html_safe
+  end
+
+  def plaintext(markdown)
+    plaintext = Redcarpet::Markdown.new(Redcarpet::Render::StripDown)
+    plaintext.render(markdown)
```
fribmendes: You don't need to assign to a variable.

app/helpers/activities_helper.rb
```diff
@@ -16,4 +19,18 @@ def can_create?
...
+  def markdown(text)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+        no_intra_emphasis: true,
+        fenced_code_blocks: true,
+        disable_indented_code_blocks: true,
+        autolink: true)
+    markdown.render(text).html_safe
+  end
+
+  def plaintext(markdown)
```
fribmendes: A better name would be `to_plaintext`, `from_markdown` or even `md_to_text`

app/helpers/activities_helper.rb
```diff
@@ -16,4 +19,18 @@ def can_create?
   def error_messages(activity)
     activity.errors.full_messages
   end
+
+  def markdown(text)
```
fribmendes: A better name would be `to_markdown` or even `render_markdown`

app/helpers/activities_helper.rb
```diff
@@ -16,4 +19,18 @@ def can_create?
...
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML,
+        no_intra_emphasis: true,
+        fenced_code_blocks: true,
```
fribmendes: Consider moving these parameters to a `markdown_config` method:

```ruby
def to_markdown(text)
  Redcarpet::Markdown.new(Redcarpet::Render::HTML, markdown_config).render(text).html_safe
end

private

def markdown_config
  {
    config: value,
    # ...
   }
end

File: app/helpers/activities_helper.rb (hunk 13 above)
Comment (ZePedroResende) [reply app/helpers/activities_helper.rb]: I spoke with @marcelo140 about this and we came to the conclusion that it would improve the readability

File: app/helpers/activities_helper.rb [hunk 17]
```diff
@@ -1,3 +1,5 @@
+require 'redcarpet/render_strip'
```
fribmendes: What happens if you don't require this?

app/helpers/activities_helper.rb
marcelo140: Yeh, but it may have been due to the indentation which I didn't perceive :S

app/helpers/activities_helper.rb
ZePedroResende: I'll get a "uninitialized constant Redcarpet::Render::StripDown" error

PR #128, opened 2017-11-01
Add partners resource

0urobor0s:
Created the partner entity and all of it's attributes


0urobor0s:
I thought that it was more simple, to only add the model
Will look on the updated issue

app/controllers/partners_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def edit
+  end
+
+  def create
+    @partner = Partner.new(partner_params)
+
+    if @partner.save
+      #redirect_to @partner
```
marcelo140: Delete this line.

app/views/partners/_form.slim
```diff
@@ -0,0 +1,42 @@
...
+    class: "partner-form-label"
+  br
+  = form.text_area :link,
+    class: "partner-form-field"
+  br
+  = form.label :logo,
+    class: "partner-form-label"
+  br
+  = form.file_field :logo,
+    class: "partner-form-field"
```
marcelo140: Just asking to be sure. Do the field's names appear capitalized?

app/views/partners/index.slim
```diff
@@ -0,0 +1,30 @@
...
+.divide40
+.container
+
+  - if can_create?
+    = link_to 'Criar Parceria', new_partner_path,
+      class: "btn btn-lg border-theme"
+  hr
+
+  - if @partners.empty?
+    h4 Parcerias não existentes
```
marcelo140: Can we change this to a friendlier message? Something like "De momento ainda não temos parcerias, volta aqui mais tarde"

app/controllers/partners_controller.rb
```diff
@@ -0,0 +1,49 @@
+class PartnersController < ApplicationController
+  before_action :set_partner, only: [:show, :edit, :update, :destroy]
```
marcelo140: You don't have a `show` method.

app/views/partners/show.slim
```diff
@@ -0,0 +1,13 @@
+.container
```
marcelo140: You don't have a `show` method.

config/routes.rb
```diff
@@ -35,11 +35,12 @@
     resources :terms,       only: [:new, :create, :destroy]
   end
 
+  resources :partners,      only: [:index, :new, :edit, :create, :update, :destroy]
```
marcelo140: You could just write `resources :partners, except: :show`

db/migrate/20171031225933_create_partners.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePartners < ActiveRecord::Migration
+  def change
+    create_table :partners do |t|
+      t.string :name, limit: 75
+      t.string :benefits
```
marcelo140: Since this attribute can hold more details than simple benefits, can we change this to description?

app/helpers/partners_helper.rb
```diff
@@ -0,0 +1,9 @@
+module PartnersHelper
+  def can_create?
+    can? :create, Partner
+  end
+
+  def error_messages(partner)
+    partner.errors.full_messages
+  end
+end
```
marcelo140: Both these methods are quite simple, I don't really think the helpers are needed.

app/controllers/partners_controller.rb
```diff
@@ -0,0 +1,46 @@
+class PartnersController < ApplicationController
+  before_action :set_partner, only: :edit
```
marcelo140: I don't understand why did you add this again. Specially for a method that you accidently deleted x)

app/views/partners/_form.slim
marcelo140: bump on this

app/controllers/partners_controller.rb
```diff
@@ -0,0 +1,46 @@
+class PartnersController < ApplicationController
```
marcelo140: You need to setup the cancancan as described in their [docs](https://github.com/CanCanCommunity/cancancan/wiki/Authorizing-controller-actions) so that the actions are correctly restricted.

app/models/ability.rb
```diff
@@ -30,5 +30,6 @@ def activity_admin_permissions(user)
     user_permissions(user)
     can :manage, Activity
     can [:index, :update], Registration
+    can :manage, Partner
```
marcelo140: Guests should be able to use the `index` action.

marcelo140:
ping @0urobor0s @ZePedroResende

app/views/partners/index.slim
```diff
@@ -0,0 +1,30 @@
...
+  hr
+
+  - if @partners.empty?
+    h4 De momento ainda não temos parcerias, volta aqui mais tarde
+
+  existentes
+  - @partners.each do |partner|
+    .row
+      - if can_create?
+        = link_to 'Editar', edit_partner_path(partner),
```
marcelo140: I don't like the edit button positioning. It moves too much content around in a way that an admin can't rely on this page to see how users see the it.

app/views/partners/index.slim
```diff
@@ -0,0 +1,30 @@
...
+          class: "btn btn-lg border-theme"
+      .col-md-2.col-sm-12
+        = link_to image_tag(partner.logo.url, class: "img-responsive activity-poster"), partner.link, target: "_blank"
```
marcelo140: When a partner has no image the layout deteriorates and instead appears an useless `missing` link. We need a different solution for this.

app/views/partners/index.slim
```diff
@@ -0,0 +1,30 @@
...
+
+      .col-md-10.col-sm-12
+        h4.partner-name
```
marcelo140: I have a feeling the logo and the text aren't correctly aligned.

app/views/partners/index.slim
```diff
@@ -0,0 +1,30 @@
...
+          = link_to partner.name, partner.link, target: "_blank"
+        ul.partner-advantages
+          .partners-list-item
+            = to_markdown(partner.description)
```
marcelo140: Both the white circle and the margin before the circles are super ugly :(

app/models/ability.rb
```diff
@@ -29,5 +30,6 @@ def activity_admin_permissions(user)
     user_permissions(user)
     can :manage, Activity
     can %i[index update], Registration
+    can :manage, Partner
```
marcelo140: We better plan how will we handle permissions.

PR #136, opened 2017-11-08
Removed "text-transform: capitalize" for h3.heading.

dinispeixoto:
"Departamento de Relações Externas e Parcerias" instead of "Departamento De Relações Externas E Parcerias". 😁

By the way, I searched for other h3.heading usages but haven't found any. So, I'm pretty sure that it's only used for department's name header in team's section.

### DIFF
--- a/app/assets/stylesheets/landing_page/_base.scss
+++ b/app/assets/stylesheets/landing_page/_base.scss
@@ -113,7 +113,6 @@ h3.heading{
     margin: 0 0 20px;
     padding-bottom: 10px;
     position: relative;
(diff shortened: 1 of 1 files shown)


martinhoaragao:
Do we want all h3.headings to not have any text transformation though? Shouldn't this be a one time thing? @dinispeixoto

marcelo140:
Personally, I'm not a fan of the capitalize transformation. It was already a problem when we were dealing with activitiy's dates and is now a problem again in the team's view. I think it's use case is way too narrow since it becomes unnatural (read 'ugly') so easily.

PR #139, opened 2017-11-15
Enhance Logout Button

9thScientist:
This PR aims to enhance the logout button on navbar's top right hand corner. 
Currently, when a user logs in, the only change visible is the navbar's top right hand corner which shows a new "logout" button. There is no real sense of which user is loged in and there's no way to get to his public profile page.
Thus, this PR will replace the logout button with the user's image. This image is a dropdown menu with two options: check the profile page and "log out".

### DIFF
--- a/app/assets/stylesheets/landing_page/_header.scss
+++ b/app/assets/stylesheets/landing_page/_header.scss
@@ -38,11 +38,25 @@
 }
 .navbar-default .navbar-nav>li>a:hover {
     color: $color-skin;
-
 }
+
+.nav .open > a, .nav .open > a:hover, .nav .open > a:focus {
+  background-color: #eeeeee;
+  border-color: #f1f1f1;
+}
+
 .navbar-brand img{
-    width:150px;
+    width: 150px;
     height: auto;
+}
+
+.navbar-user {
+    height: 50px;
+}
+
+.navbar-signout {
+    padding: 0px;
+    margin-left: 20px;
 }
 
 .btn-login {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/landing_page/_header.scss
```diff
@@ -38,13 +38,27 @@
...
+}
+
+.navbar-signout {
+    padding: 0px;
+    margin-left: 20px;
+}
+
 .btn-login {
   color: $color-skin !important;
   font-weight: bold !important;
```
marcelo140: Not sure, but I think `!important` is a bad practice. @martinhoaragao ?

app/views/layouts/lp/_navbar.html.slim
```diff
@@ -28,7 +28,16 @@
           = link_to 'CONTACTOS', contact_path
         - if current_user
           li
-            = link_to 'Log Out', sign_out_path, method: :delete
+            .dropdown.show
+              = link_to image_tag(current_user.image, class: 'img-circle navbar-user'), '#',
```
marcelo140: I think this would look nicer if you broke the arguments to multiple lines. What do you think about a partial for the entire dropdown menu?

app/assets/stylesheets/landing_page/_header.scss
martinhoaragao: Oh yes, it is pretty damn bad. 

You probably need to be more specific like saying .general .btn-login . It's most likely not applying because there's a rule which is more specific than the one you just created.

marcelo140:
ping @9thScientist

app/assets/stylesheets/landing_page/_header.scss
9thScientist: I was not the one who wrote that. Since we are on a mission to get things done, I'll leave this for another time.

Issue #129, opened 2017-11-02
Translation bug in will_paginate


9thScientist:
If you go to [previous activities page](http://cesium.di.uminho.pt/activities?show=previous) you will encounter a visual bug in will_paginate. 
![](https://screenshots.firefoxusercontent.com/images/a4ee62a7-1cd4-402d-86ea-078a84f414f1.png)
If you hover the `Next Label` button, a tooltip will popup that reads:

> translation missing: pt.will_paginate.next_label

I'm assuming it's a problem with translation, but do we really need it to be written out? Can't we use a [font awesome icon](http://fontawesome.io/icon/chevron-right/)?

marcelo140:
The problem is that I setup `pt` as the default locale in a previous PR. This could be easily fixed by adding a translation for `next label` and `previous label`.

PR #140, opened 2017-11-15
Update ruby and rails

marcelo140:
The current ruby version is quite difficult to install (problems with docs and openssl) and gives me segmentation fault when I try to change the CSS. Since rails 4 has some problems with ruby 2.4 (Fixnum was deprecated), I ended up bumping the rails version too.

I tested in staging all our common operations (login, logout, create & edit activities, registrate, unregistrate, confirm/cancel confirmation) and, after some changes, I made it work. Still, the fact that we have no tests doesn't give me much confidence about this PR :S.


fribmendes:
Before I review this, two questions:

1. Do you want to wait for specs?
2. Did you follow an upgrade guide? There are things that are now assumed as default in Rails 5, in terms of project directory.

marcelo140:
1. I don't think the app would break, but I do prefer to wait for the specs since (hopefully) they're not that far away.

2. While I was upgrading I actually passed by the upgrade guide but forgot to stick to it. Thank you for reminding me.

app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ApplicationRecord
```
fribmendes: ⚠️

marcelo140:
Done. Fixes https://github.com/cesium/atomic/issues/129 and https://github.com/cesium/atomic/issues/147. @fribmendes will you want to review this or do you prefer a let's just break it approach? I'm all for it :)

fribmendes:
Let's just break it 😉

PR #144, opened 2017-12-06
It should be .env.sample not .env.example

fribmendes:


fribmendes:
@marcelo140 ping on this.

PR #145, opened 2017-12-06
Add activities controller specs

fribmendes:


fribmendes:
@marcelo140 here you go. Also, you guys should set up travis or something

PR #149, opened 2018-04-15
Add news

### LINKED ISSUE #146: Dropdown list doesn't work when you are on the page "news"
when you are using a phone and the navbar becomes a dropdown list, you can't open the dropdown list if you are on the page news

machadovilaca:
The main goal was to replace the usage of wordpress for CeSIUM news'.

This feature follows very closely the approach used on the activities section, right now it even shares posting permissions, this is, whoever can manage activities can manage news.

Fixes #146


db/migrate/20180414130746_create_news.rb
```diff
@@ -0,0 +1,12 @@
+class CreateNews < ActiveRecord::Migration
```
marcelo140: Due to the recent migration to rails 5.1, you need to specify the migration version. In this case `ActiveRecord::Migration[4.2]`

app/controllers/news_controller.rb
```diff
@@ -0,0 +1,56 @@
+class NewsController < ApplicationController
```
marcelo140: I don't like the fact that we can't distinguish a single piece of `news` from multiple `news` x/

app/models/ability.rb
```diff
@@ -29,5 +29,6 @@ def activity_admin_permissions(user)
     user_permissions(user)
     can :manage, Activity
     can %i[index update], Registration
+    can :manage, News
```
marcelo140: I think I prefer the creation of a `news_admin` or something instead of postponing the issue.

app/controllers/news_controller.rb
marcelo140: @xicon73 suggested articles

machadovilaca:
@marcelo140 I've added requested changes

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,58 @@
+class ArticlesController < ApplicationController
+  load_and_authorize_resource
+
+  def index
+    @articles = Article.all.order("created_at DESC").paginate(page: params[:page], per_page: 5)
```
ZePedroResende: 80 chars max

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    if @article.save
+      redirect_to @article
+    else
+      render :new
+    end
+  end
+
+  def edit
+    @article = Article.find_by(id: params[:id])
```
ZePedroResende: Find_by is unnecessary here a find is fine

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def update
+    @article = Article.find_by(id: params[:id])
```
ZePedroResende: same as above

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    if @article
+      update_article(article_params)
+    else
+      not_found
+    end
+  end
+
+  def destroy
+    Article.find_by(id: params[:id]).try(:destroy)
```
ZePedroResende: same as above

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    redirect_to articles_path
+  end
+
+  private
+
+  def article_params
+    params.require(:article).permit(:name, :text, :poster)
+  end
+
+  def update_article(params)
```
ZePedroResende: We should not use non RESTful methods in here

app/models/ability.rb
```diff
@@ -30,4 +33,14 @@ def activity_admin_permissions(user)
     can :manage, Activity
     can %i[index update], Registration
   end
+
+  def article_admin_permissions(user)
+    user_permissions(user)
+    can :manage, Article
+  end
+
+  def full_admin_permissions(user)
```
ZePedroResende: Is this necessary ?

config/database.yml
```diff
@@ -12,9 +12,9 @@ default: &default
 development:
   <<: *default
   encoding: unicode
-  database: atomic_development
+  database: atomic
   username: postgres
-  password:
+  password: postgres
```
ZePedroResende: same

config/database.yml
```diff
@@ -12,9 +12,9 @@ default: &default
 development:
   <<: *default
   encoding: unicode
-  database: atomic_development
+  database: atomic
```
ZePedroResende: what is this ?

config/initializers/omniauth.rb
```diff
@@ -1,4 +1,4 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
-  provider :github, ENV['GITHUB_APP_ID'], ENV['GITHUB_APP_SECRET']
+  provider :github, 'b97c6aeafe77999aee8e', '3ac7f187a76847d218bc4e0d621106b4fda2fb39'
```
ZePedroResende: Hello from vilaça account !

app/controllers/articles_controller.rb
machadovilaca: "find_by" returns nil if the requested article does not exist in the database, on the other hand "find" will return an error, if we use find both in update and destroy we will need to handle the errors using the rescue block

app/controllers/articles_controller.rb
machadovilaca: "find_by" returns nil if the requested article does not exist in the database, on the other hand "find" will return an error, if we use find both in this, update and destroy methods we will need to handle the errors using the rescue block

0urobor0s: I think we should use `find`, also the error is already caught and 404s

app/models/ability.rb
```diff
@@ -32,4 +33,14 @@ def activity_admin_permissions(user)
...
+
+  def article_admin_permissions(user)
+    user_permissions(user)
+    can :manage, Article
+  end
+
+  def admin_permissions(user)
+    article_admin_permissions(user)
+    activity_admin_permissions(user)
+  end
```
marcelo140: Merge all admin permissions together.

Issue #108, opened 2017-10-22
Pages lack meta tags for the open graph protocol


martinhoaragao:
All pages but the activities' are lacking the open graph protocol meta tags on the head.

Using either a picture from the carousel or one of CeSIUM's logo wallpapers might be best.

9thScientist:
We first need to know if we are going to use CeSIUM's logo or the carousel's first image.
I think for most cases the carousel image will be the best, since it helps to spread the current most important event.
Although I understand why we would use one of the CeSIUM's logo wallpaper, as it's a more generic image.

martinhoaragao:
There are 2 issues in my opinion with using the first image in the carousel:
+ Related to implementation, currently it's just a manually defined image, and not a resource correct?
+ Scrapper's cache. We can't force all websites that have scrapped our open graph tags to refresh them. As such during SEI'18 we might still show CodeWeek 2017 when we link our page on facebook or Slack.

As such I do think a more generic view that has the same relevance now or in a few months is a wiser choice.

PR #162, opened 2018-04-30
Adds open graph meta tags to most pages

9thScientist:
This PR is intended to fix issue #108 and add open graph protocol meta tags to the following pages:
 - Index
 - About
 - Team
 - Partners
 - Activities
 - Contacts


martinhoaragao:
Considering these are meta tags per page, might as well throw in more info, such as a specific description per page. Why not add meta tags that are set by default for all pages when there's no explicit ones set?

According to [Open Graph Protocol page](http://ogp.me/) , `og:url` and `og:type` are basic metadata that should be added.  I'd recommend throwing in a `og:site_name` as well.

app/views/welcome/team.html.slim
```diff
@@ -1,3 +1,10 @@
+- content_for :head do
+  meta property='og:image' content='#{image_url("cesium-thumb.png")}'
+  meta property='og:title' content='Equipa'
+  meta property='og:site_name' content='CeSIUM'
+  meta property='og:type' content='website'
+  meta property='og:url' content='http://cesium.di.uminho.pt/team'
+  meta property='og:description' content='O CeSIUM é um Centro de estudantes de Engenharia Informática da Universidade do Minho composto por alunos voluntários que têm como objectivo principal representar e promover o curso. É um núcleo unido e dinâmico capaz de proporcionar experiências únicas e enriquecedoras para a tua futura vida profissional.'
```
nelsonmestevao: Change to https, please 🙏

nelsonmestevao (approved):

PR #165, opened 2018-05-01
Add limit to participants

ZePedroResende:


app/models/activity.rb
```diff
@@ -39,6 +40,10 @@ def registered?(user)
     registrations.find_by(user_id: user.id)
   end
 
+  def can_register?
+    participants.count < number_participants
+  end
```
9thScientist: Shouldn't this be a helper method?

app/views/activities/_form.slim
```diff
@@ -30,10 +30,11 @@
   = form.text_field :speaker,
     class: "activity-form-field"
 
-  = form.label :total_rating,
+  = form.label :number_participants,
     class: "activity-form-label"
   br
-  = form.text_field :total_rating,
+  = form.number_field :number_participants,
+    value: 20,
```
9thScientist: You are setting by default, the maximum number of participants as 20. But what if we don't want to specify the maximum number of participants?

config/locales/pt.yml
```diff
@@ -25,3 +25,12 @@ pt:
...
     next_label: "Seguinte"
+
+  activerecord:
+    errors:
+      models:
+        activity:
+          attributes:
+            number_participants:
+              blank: "vazio"
+              not_a_number: "nao e um numero"
```
9thScientist: Please, respect the language of Camões.

app/views/activities/_form.slim
ZePedroResende: You can give 0 to have no limite

config/locales/pt.yml
ZePedroResende: :'( 9.5 in Portuguese

app/models/activity.rb
```diff
@@ -39,6 +40,10 @@ def registered?(user)
     registrations.find_by(user_id: user.id)
   end
 
+  def can_register?
```
marcelo140: The name seems too vague for me. It should be something more related to the fact that the talk isn't full. Something along the lines `has_sits_available?`

app/views/activities/_form.slim
marcelo140: I don't like this neither. If it's unlimited, it should be left blank.

app/controllers/activities/registrations_controller.rb
```diff
@@ -3,8 +3,10 @@ class Activities::RegistrationsController < ApplicationController
   load_and_authorize_resource
 
   def create
-    unless @activity.registered?(current_user)
-      Registration.create(activity_id: @activity.id, user_id: current_user.id)
+    Activity.transaction do
+      if !@activity.registered?(current_user) && @activity.can_register?
+        Registration.create(activity_id: @activity.id, user_id: current_user.id)
+      end
```
marcelo140: This should be a method in the model.

ZePedroResende:
@9thScientist  @marcelo140  ping, i think its all done

PR #166, opened 2018-05-01
Refactor users and identities

marcelo140:
- Renames users to identities
- Renames members to users
- Remove user details from identities
- Update user details
- Change everything else to fit in.

First steps for https://github.com/cesium/atomic/pull/166.


app/models/user.rb
```diff
@@ -1,6 +1,12 @@
 class User < ApplicationRecord
   has_many :registrations
-  belongs_to :member, optional: true
+  has_one :identity
```
ZePedroResende: shouldn't this be has_many :identity ? so a user can use diferent providers ?

marcelo140:
@ZePedroResende sad

ZePedroResende:
@marcelo140 :sob:

PR #168, opened 2018-05-04
fix nil speaker error

### LINKED ISSUE #167: Speaker in activities aren't implemented correctly
Currently, `speaker` allows `nil` values. Unfortunetelly, `to_markdown` and `to_plaintext` values can't handle `nil` values and cause `500: Internal Server Error`.

machadovilaca:
Couldn't reproduce the issue but I think this fixes #167

### DIFF
--- a/app/views/shared/activity/_description_date.slim
+++ b/app/views/shared/activity/_description_date.slim
@@ -1,9 +1,11 @@
 p.lead.activity-description
   = truncate(to_plaintext(activity.description), length: 140)
 
-- if activity.speaker
-  i.far.fa-user
-    = " " + to_plaintext(activity.speaker)
+- if activity.speaker?
+  ul.list-inline
+    li
+      i.far.fa-user
+    = to_plaintext(activity.speaker)
 
 ul.list-inline
   - if activity.start_date.to_date == activity.end_date.to_date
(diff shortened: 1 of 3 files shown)


marcelo140:
One more thing, currently the speaker feels like it's part of the activity's description (both in index and show). What do you think of highlighting it by putting it on info panel?

machadovilaca:
@marcelo140 How does it look?
index
show

marcelo140:
I'm ok with `show` but I still don't like it's positioning on the `index`. Can we just not show the speaker on the `index`? @martinhoaragao @dinispeixoto opinions on this ?

app/views/shared/activity/_description_date.slim
```diff
@@ -1,9 +1,10 @@
-p.lead.activity-description
-  = to_plaintext(activity.speaker)
-
 p.lead.activity-description
   = truncate(to_plaintext(activity.description), length: 140)
 
+- if activity.speaker
+  i.far.fa-user
+    = " " + to_plaintext(activity.speaker)
```
ZePedroResende: this doesn't add an empty space ?

machadovilaca: that's the objective, anyway if the "look" is approved I will change it to an acceptable way to add space between the icon and the name

ZePedroResende: Ok that's what i was gonna sugest . good job

ZePedroResende:
@machadovilaca  ping

machadovilaca:
@ZePedroResende waiting for views approval

ZePedroResende:
@machadovilaca  Approved

PR #172, opened 2018-10-03
Create a file for humans too

nelsonmestevao:
We already have a file for robots. What about one for us? That's what
this pull request tries to fix.

Read more about this [here](http://humanstxt.org/).


ZePedroResende:
The logo looks kinda stretched 😢

public/humans.txt
```diff
@@ -0,0 +1,36 @@
+
```
ZePedroResende: stretched

public/humans.txt
```diff
@@ -0,0 +1,36 @@
...
+                                  `.
+                               `-/++/:-`
```
ZePedroResende: The logo looks kinda stretched 😢

nelsonmestevao: I made an update.

PR #175, opened 2018-10-19
Updated photos of the team

FranciscoLira:
Just the photo files were updated


ZePedroResende:
@FranciscoLira the standard name for a branch is (your name acronym)-(name of branch). 
Also please make a single commit with the photos json branch. There's no need for two commits that have one purpose

PR #179, opened 2018-10-19
Add limit to participants

ZePedroResende:


app/views/shared/activity/_panel.slim
```diff
@@ -24,38 +24,36 @@
...
-    .col-md-12
-
+    .col-md-6
       = render partial: "shared/activity/info",
         locals: { icon: "coffee", text: "Coffee Break" }
 
-  // Rating
-  - if @activity.total_rating != nil
-    .col-md-12
+  - if @activity.limit_number_participants
```
machadovilaca: This should not be displayed if activity does not allow registrations (`allows_registrations` is false).

app/views/activities/_form.slim
```diff
@@ -30,12 +30,19 @@
   = form.text_field :speaker,
     class: "activity-form-field"
 
-  = form.label :total_rating,
+  = form.label :limit_number_participants,
```
machadovilaca: Move this to under the `:allows_registrations` (to line 99) part of the form and do the same as you did for the `#is-number-participants-limited`, just show this if activity allows registrations.

app/views/activities/_form.slim
```diff
@@ -30,12 +30,19 @@
...
     class: "activity-form-label"
   br
-  = form.text_field :total_rating,
+  = form.check_box :limit_number_participants,
     class: "activity-form-field"
-
+  br
+  div#is-number-participants-limited hidden=""
```
machadovilaca: If I open the update form or reload the page on an activity with limited number of participants, the checkbox is selected but the `#is-number-participants-limited` is hidden. You need to verify this in your js.

PR #182, opened 2018-11-21
Fix footer when content isn't big enough

nelsonmestevao:
Prior to this change, when the content was really small (or even empty), footer would be not on the bottom of the page.
 
Read more [here](https://stackoverflow.com/questions/27985248).


nelsonmestevao:
Fix footer at the bottom with small content :tada: 

I tried to follow [this](https://matthewjamestaylor.com/bottom-footer)
but there were too many demanding changes.

Luckily, [stackoverflow](https://stackoverflow.com/questions/643879/css-to-make-html-page-footer-stay-at-bottom-of-the-page-with-a-minimum-height-b) exists.

db/schema.rb
```diff
@@ -88,7 +88,6 @@
     t.string "student_id", default: ""
     t.boolean "is_buddy"
     t.boolean "admin", default: false, null: false
-    t.boolean "activity_admin", default: false, null: false
```
ZePedroResende: This seem strange, maybe its better you checkout the schema or something

nelsonmestevao: Well, I just run `bin/setup`. I found odd but have no idea. Shouldn't keep it then?

Gemfile.lock
```diff
@@ -1,5 +1,5 @@
 GIT
-  remote: git://github.com/capistrano/rbenv.git
+  remote: https://github.com/capistrano/rbenv.git
```
ZePedroResende: Why ?

db/schema.rb
ZePedroResende: you where changing the current database schema without a migration so there would be problems :)

Gemfile.lock
nelsonmestevao: Just `bundle install` :man_shrugging: (but everything worked perfectly)

db/schema.rb
nelsonmestevao: Is the setup process wrong?

nelsonmestevao:
@ZePedroResende just before I merge this changes, should I checkout the `Gemfile.lock` or not?

db/schema.rb
ZePedroResende: No, i remember someone else had this problem i think . 
I'm not really familiar with the codebase at this point.
This is in fact weird. @herulume you should check this when possible !

ZePedroResende:
> @ZePedroResende just before I merge these changes, should I checkout the `Gemfile.lock` or not?

We have the git port so we can use git instead of https for the gems, although there's nothing wrong with using https at the top of my head. 
Your choice i think.

ZePedroResende:
@nelsonmestevao @herulume  read [this](http://hjhart.blog/2017/04/05/using-bundler-to-securely-install-your-private-github-gems.html).
I didn't had this view. Maybe something you could do in the next pr

db/schema.rb
herulume: @ZePedroResende Yes sir, added to my bullet journal.

PR #190, opened 2019-11-05
Categorize display of team members

miguelbrandao:
As requested by @FranciscoLira


herulume:
> @herulume we already have them, the intent is to change the looks in one PR and the team data in another one

No PR was opened with that afaik.

PR #192, opened 2019-11-06
Add new team

miguelbrandao:


miguelbrandao:
> I just don't like the fact that "Presidência" is broken in two lines. The only solution I guess would be to change margins...

@nelsonmestevao I agree, but we should do that in another PR, maybe in #190

PR #209, opened 2020-11-15
Add docker-compose for database

machadovilaca:


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.1'
+
+services:
+
+  atomic_db:
+    image: postgres
+    restart: always
+    environment:
+      POSTGRES_PASSWORD: atomic
```
nelsonmestevao, line 9: I think it would be better to use `${DB_PASSWORD}` and `${DB_USERNAME}` in order to be in sync with the `.env` file.

nelsonmestevao:
What do you think about mention this option in order to setup the database on the REAME?

nelsonmestevao (approved):
:rocket: with :whale:

PR #221, opened 2021-03-06
Rename "Notícias" to CeSIUMnology

nelsonmestevao:


machadovilaca (changes requested):
Also change [here](https://github.com/cesium/atomic/blob/master/app/views/articles/_articles_breadcrumb.slim), line 6 and 11

machadovilaca (approved):

PR #222, opened 2021-03-07
Fix test environment on local machine

nelsonmestevao:
Closes #215.


docker-compose.yml
```diff
@@ -1,12 +1,20 @@
 version: '3.1'
 
 services:
-
-  atomic_db:
-    image: postgres
-    restart: always
+  db:
+    image: postgres:9.4
+    container_name: atomic_db
```
machadovilaca, line 6: Why change the name from `atomic_db` to `db` and then explicitly set the container name to `atomic_db`?

machadovilaca (approved):

docker-compose.yml
nelsonmestevao: db is the service, is easier to write docker-compose db run (etc) than docker-compose atomic_db run.

PR #241, opened 2022-01-21
Revert to Ruby 2.7.5 to maintain 'Paperclip' dependency

miguelbrandao:
`Paperclip` was used to store and manage uploaded images before Rail's Active Storage came along.

Turns out, `Paperclip` was deprecated shortly after the release of Active Storage and it does not support Ruby 3.

Migrating to our records to Active Storage is the long term solution, but the documentation is sparse and I've been at it for a few hours with moderate success, but unfortunately not enough.

So, for the time being, let's revert to Ruby 2.7.5 so we have a functioning site and think about migrations later.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-ruby 3.0.3
+ruby 2.7.5
 nodejs 16.13.0
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
+ruby '2.7.5'
+
 source "https://rubygems.org"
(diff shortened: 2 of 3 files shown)


miguelbrandao:
If anyone wants to try migrating away from `Paperclip` here you go: https://github.com/thoughtbot/paperclip/blob/master/MIGRATING.md

And a little help for a bug you'll find: https://stackoverflow.com/questions/54333936/undefined-method-execute-prepared-rails-paperclip-to-activestorage-migration

Pray to whatever god or entity you believe in

nelsonmestevao (approved):
Very sad :/

PR #244, opened 2022-04-09
Update partnerships

PietroPan:
Remove old partnerships and their respective images
Add new partnerships and their respective images


nelsonmestevao (approved):
Well done :clap:

PR #248, opened 2022-07-07
The other half of Slim to Html

zemmartins:


miguelbrandao (changes requested):
Remove the comments and slim controlls

ruioliveira02:
We are refactoring this in Elixir, so this PR is no longer useful. Thank you for contributing :pray:

PR #252, opened 2022-10-22
Add activities

feliciofilipe:


matildeopbravo:
Shouldn't activities contain more information, such as location, date and time?

feliciofilipe:
> Shouldn't activities contain more information, such as location, date and time?

Yeah, I was thinking on doing through a one to many relation with a sessions table. By doing so we are able to have several moments ("sessions") with different times/locations per activity, if we wish so.

Maybe is better if I'll add it in this PR :pray:

MarioRodrigues10 (approved):

matildeopbravo (approved):

PR #253, opened 2022-11-07
Add enrollments

feliciofilipe:


test/atomic/activities_test.exs
```diff
@@ -129,4 +129,60 @@ defmodule Atomic.ActivitiesTest do
       assert %Ecto.Changeset{} = Activities.change_session(session)
     end
   end
+
+  describe "enrollments" do
+    alias Atomic.Activities.Enrollment
+
+    import Atomic.ActivitiesFixtures
```
ruioliveira02, line 136: We should probably start using ```ex_machina``` for testing straight-away

lib/atomic/activities.ex
```diff
@@ -200,4 +201,146 @@ defmodule Atomic.Activities do
...
+
+  ## Examples
+
+      iex> list_enrollments()
+      [%Enrollment{}, ...]
+
+  """
+  def list_enrollments do
+    Repo.all(Enrollment)
+  end
```
ruioliveira02, line 218: Wouldn't it be useful to have functions to list enrollments filtered by user and activity? Allowing to preload the associations by adding one parameter could also prove useful

lib/atomic/activities.ex
```diff
@@ -200,4 +201,146 @@ defmodule Atomic.Activities do
...
+      %Enrollment{}
+
+      iex> get_enrollment!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_enrollment!(id), do: Repo.get!(Enrollment, id)
+
+  @doc """
+  Creates a enrollment.
```
ruioliveira02, line 237:
```suggestion
  Creates an enrollment.
```

Same for update and delete

ruioliveira02:
There should be seeds for activities in general

PR #254, opened 2022-11-19
Add departments

feliciofilipe:


lib/atomic/departments/department.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  @required_fields ~w(name)a
+
+  @optional_fields []
+
+  schema "departments" do
+    field :name, :string
+
+    has_many :activities, Activity
+
```
danielsp45, line 17: What do you think about the Department table having a `has_many` parameter for Users that are members of that department?

feliciofilipe: I was thinking that.

It should be a `many_to_many` relation, since a person can be a member of more than one department.

With that being said, departments change throughout the years (they appear/disappear and change names) so I want to take a set back a think which is the most sophisticated way of doing this relation so that we do not lose any of the history when it comes to the roles/contribution that people have/had in CeSIUM.

I will do this in a future PR :pray::pray:

danielsp45: Yes you're totally right, I didn't even take that into consideration 🙏🏻

PR #266, opened 2023-02-19
Fix departments form

feliciofilipe:


MarioRodrigues10 (dismissed):
The problem is in ```lib/atomic_web/live/department_live/show.ex``` in this line:

```c
     |> assign(:department, Departments.get_department!(id, preloads: :activities))}
```

Please find another way to ```preload activities```

PR #280, opened 2023-04-11
Setup mailer

MarioRodrigues10:


config/dev.exs
```diff
@@ -28,7 +28,8 @@ config :atomic, AtomicWeb.Endpoint,
     # Start the esbuild watcher by calling Esbuild.install_and_run(:default, args)
     esbuild: {Esbuild, :install_and_run, [:default, ~w(--sourcemap=inline --watch)]},
     tailwind: {Tailwind, :install_and_run, [:default, ~w(--watch)]}
-  ]
+  ],
+  frontend_url: "http://localhost:4000"
```
ruioliveira02, line 32: If this is configured in `.env.dev`, why don't you fetch it from there and only use this as a default value?

config/runtime.exs
```diff
@@ -63,6 +63,12 @@ if config_env() == :prod do
     ],
     secret_key_base: secret_key_base
 
+  config :atomic, AtomicWeb.Endpoint, server: true
+
+  config :atomic, Atomic.Mailer,
+    adapter: Swoosh.Adapters.Sendgrid,
```
ruioliveira02, line 69:
```suggestion
    adapter: Swoosh.Adapters.Sendinblue,
```

It is not certain we will use Sendinblue (especially as we may reach a time where we send > 300 emails / day), but for now I think it is the best choice

lib/atomic_web/views/helpers.ex
```diff
@@ -0,0 +1,32 @@
+defmodule AtomicWeb.Helpers do
```
RuiL1904, line 1: I've made a major update to this module [on this PR](https://github.com/coderdojobraga/bokken/pull/168).
It would be nice if you use it here 👀

ruioliveira02 (approved):

PR #281, opened 2023-04-14
Add certificate generation for activities

ruioliveira02:
Notes before merging:
- Tailwind is not working in email templates
- Seeds will need to be updated to add more variety to times in activities
- ~~I still want to document the module~~
- There is no styling on emails and PDFs: this is on purpose and will be added in another PR
- ~~History is messed up because I based this branch off of `mr/setup-mailer` which hasn't been merged yet~~

To test:
- Go to `config/config.exs` and change the `@daily` with `"* * * * *"`. This will make the task run every minute.

How does this work:
- This task will run at midnight every day (the choice of midnight - and this will be documented in the code - is because no activity would ever go beyond midnight: most would be from 9am:6pm)
- As a result, the task will search for all activities which ended (by ended we mean their last session ended) within the last 24 hours (as the task runs daily)


config/config.exs
```diff
@@ -67,6 +67,12 @@ config :tailwind,
     cd: Path.expand("../assets", __DIR__)
   ]
 
+config :atomic, Atomic.Scheduler,
+  jobs: [
+    # Runs every midnight:
```
MarioRodrigues10, line 72: Which midgnight? Lisbon UTC?

lib/atomic_web/emails/activity_emails.ex
```diff
@@ -0,0 +1,21 @@
...
+    |> assign(:enrollment, enrollment)
+    |> attachment(certificate)
+    |> render_body("activity_certificate.html")
+  end
+
+  defp base_email(to: email) do
+    new()
+    |> from({"Atomic", "noreply@atomic.cesium.link"})
+    |> to(email)
+    |> reply_to("caos@cesium.di.uminho.pt")
```
MarioRodrigues10, line 19:
```suggestion
    |> reply_to("support@atomic.cesium.pt")
```

lib/atomic_web/templates/email/activity_certificate.html.eex
```diff
@@ -0,0 +1,20 @@
...
+   <head>
+      <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+      <meta http-equiv="Content-Type" content="text/html charset=UTF-8" />
+   </head>
+   <body>
+        <div>
+            <p>Bom dia <%= @enrollment.user.name %>,</p>
+
+            <p>Gostávamos de agradecer a tua participação na atividade <%= @enrollment.activity.title %>. Enviamos, em anexo,
+            o certificado de participação. Alternativamente, podes consultar o certificado no teu perfil da plataforma Atomic.
```
MarioRodrigues10, line 12:
```suggestion
            <p>Gostávamos de agradecer a tua participação na atividade <%= @enrollment.activity.title %>. Enviamos, em anexo,
            o certificado de participação. Alternativamente, podes consultar o certificado no teu perfil.
```

We should not mention Atomic every time, we could have Atomic logo in the feature but there's no need to be mentioning it all the time.

lib/atomic_web/templates/email/activity_certificate.html.eex
```diff
@@ -0,0 +1,20 @@
...
+            </p>
+
+            <p>Esperamos encontrar-te em atividades futuras!</p>
+
+            <p>Este email foi gerado automaticamente. Por favor não respondas para este endereço de email.</p>
```
MarioRodrigues10, line 17:
```suggestion
            <p>Este email foi gerado automaticamente. Por favor não respondas para este endereço de email, para alguma dúvida envia email para <%= organization.email %> .</p>
```

MarioRodrigues10 (changes requested):
Please work on these small things

config/config.exs
ruioliveira02: Whenever midnight is in the server's timezone. Meaning, it will most likely be in UTC

lib/atomic_web/templates/email/activity_certificate.html.eex
ruioliveira02: The problem is getting the organization will be a PITA. I was hoping to be able to trap someone else to do it

ruioliveira02: Plus I don't think that with the `reply_to` set it makes sense to include that text.

MarioRodrigues10: Well , users report to organizations and organizations report to us if there is any problem. At least that's what i think that should happen.

ruioliveira02: Sure. I still don't think we should be too hung up on this, as this is not the version that will be in production and is just a placeholder

lib/atomic/quantum/certificate_delivery.ex
```diff
@@ -0,0 +1,131 @@
...
+
+    sq =
+      from s in Session,
+        group_by: [s.activity_id],
+        select: %{finish: max(s.finish), activity_id: s.activity_id}
+
+    from s in subquery(sq),
+      where: s.finish >= ^minimum_finish and s.finish <= ^now,
+      select: s.activity_id
+  end
```
MarioRodrigues10:
```suggestion
  defp last_sessions_query() do
    now = DateTime.utc_now()
    minimum_finish = DateTime.add(now, -24, :hour)

    sq =
      from s in Session,
        where: s.finish >= ^minimum_finish and s.finish <= ^now,
        group_by: [s.activity_id],
        select: %{finish: max(s.finish), activity_id: s.activity_id}

    from s in subquery(sq),
      select: s.activity_id
  end
```

lib/atomic/quantum/certificate_delivery.ex
```diff
@@ -0,0 +1,131 @@
...
+  24 hours ago.
+
+  Activities are said to end when their last session finishes. Only
+  participants who were present are eligible to receive certificates
+
+  """
+  def send_certificates do
+    enrollments = included_enrollments()
+
+    for enrollment <- enrollments do
```
MarioRodrigues10:
```suggestion
    included_enrollments()
    |> Enum.each(fn enrollment -> 
```

MarioRodrigues10 (changes requested):

priv/repo/seeds/activities.exs
MarioRodrigues10, line 1: Please add a %Location{} for each ActivitySession

MarioRodrigues10 (changes requested):

MarioRodrigues10 (approved):

PR #284, opened 2023-06-08
Add events and qrcodes to activities

MarioRodrigues10:


lib/atomic/activities.ex
```diff
@@ -233,6 +233,27 @@ defmodule Atomic.Activities do
...
+      |> Repo.one()
+
+    if enrollment do
+      Repo.delete(enrollment)
+      {:ok, "Participation canceled"}
+    else
+      create_enrollment(activity, user)
+      {:ok, "Participation confirmed"}
+    end
+  end
```
ruioliveira02:
```suggestion
  def switch_user_enrolled(user, activity) do
    enrollment =
      Enrollment
      |> where(user_id: ^user.id, activity_id: ^activity.id)
      |> Repo.one()

    if enrollment do
      Repo.delete(enrollment)
      :cancelled
    else
      create_enrollment(activity, user)
      :enrolled
    end
  end
```

The name as it is is pretty confusing (I am not particularly happy with my suggestion either). Regardless, we should document this function to make sure its functionality is understood.

Also, if you are always returning `{:ok, something}`, might as well just return the `something`, as this will never fail

lib/atomic/activities/activity.ex
```diff
@@ -12,10 +12,11 @@ defmodule Atomic.Activities.Activity do
   alias Atomic.Activities.Speaker
   alias Atomic.Departments
   alias Atomic.Departments.Department
+  alias Atomic.Events.Event
   @required_fields ~w(title description
                     minimum_entries maximum_entries)a
 
-  @optional_fields []
+  @optional_fields ~w(event_id)a
```
ruioliveira02, line 19: Why is this field optional?

lib/atomic_web/controllers/enrollments_controller.ex
```diff
@@ -0,0 +1,29 @@
+defmodule AtomicWeb.EnrollmentController do
```
ruioliveira02: Why is this is a controller and not part of the `live`?

ruioliveira02 (commented):
We should implement a system similar to badges in SEI. Now that I am thinking, adding gamification to our events (mainly recreational ones) might be worth looking into

lib/atomic/activities/activity.ex
MarioRodrigues10: An activity could belong to an event or not

lib/atomic_web/controllers/enrollments_controller.ex
MarioRodrigues10: Well we might need to create an API in the future, that's why i'm using controllers, but i can change it to live if you consider it better.

lib/atomic/activities.ex
MarioRodrigues10: You're right, thanks !!

MarioRodrigues10:
> We should implement a system similar to badges in SEI. Now that I am thinking, adding gamification to our events (mainly recreational ones) might be worth looking into

We already have our own platforms to  events, i think we shouldn't implement a badge system for now.

lib/atomic_web/controllers/enrollments_controller.ex
```diff
@@ -0,0 +1,29 @@
...
+    user = Repo.get!(User, user_id)
+    activity = Repo.get!(Activity, activity_id)
+    enrollment = Activities.get_enrollment(activity, user)
+
+    if enrollment do
+      Activities.update_enrollment(enrollment, %{present: true})
+
+      conn
+      |> put_flash(:info, "Participation confirmed")
+      |> redirect(to: "/activities/#{activity.id}")
```
ruilopesm:
```suggestion
      |> redirect(to: ~p"/activities/#{activity.id}")
```

> Using ~p for route paths ensures our application paths and URLs stay up to date with the router definitions. The compiler will catch bugs for us, and let us know when we change routes that are referenced elsewhere in our application.

Source: https://hexdocs.pm/phoenix/routing.html

lib/atomic_web/controllers/enrollments_controller.ex
```diff
@@ -0,0 +1,29 @@
...
+    else
+      conn
+      |> put_flash(:error, "Enrollment not found")
+      |> redirect(to: "/activities/#{activity.id}")
```
ruilopesm: Same here

lib/atomic_web/controllers/enrollments_controller.ex
ruioliveira02: Shouldn't we use `Routes.activity_show_path/3` here?

MarioRodrigues10:
In case you're wondering why there's changes on 47 files, i did that in order to fix lint.

lib/atomic_web/live/scanner_live/index.ex
```diff
@@ -0,0 +1,23 @@
...
+    {:noreply,
+     socket
+     |> assign(:current_page, :scanner)
+     |> assign(:title, "Scanner")}
+  end
+
+  @impl true
+  def handle_event("scan", pathname, socket) do
+    {:noreply, push_redirect(socket, to: "/#{pathname}")}
+  end
```
ruioliveira02: Can we do this without redirecting to another page? I feel that flow (we use in SEI) is better than JOIN in that regard. It allows for much faster scanning

MarioRodrigues10: Well how do you pretend to go to confirmation page then?

lib/atomic_web/live/redeem_live/form_component.html.heex
```diff
@@ -0,0 +1,61 @@
+<div>
+  <.form let={_f} for={:activity_user} id="activity_user-form" phx-target={@myself} phx-submit="save">
+    <div class="min-h-full bg-white border-r border-gray-200">
+      <div class="pt-5 pr-6 pb-5 pl-4 border-gray-200 sm:pl-6 lg:pl-8 xl:pt-6 xl:pl-6 border-y">
+        <div class="flex justify-between items-center h-10">
+          <h1 class="flex-1 text-lg font-medium">Task</h1>
```
joaodiaslobo:
```suggestion
          <h1 class="flex-1 text-lg font-medium">Activity</h1>
```

lib/atomic_web/live/scanner_live/index.ex
ruioliveira02: That is the neat part, you won't. I don't believe we have a need for that step in this workflow.

ruioliveira02 (approved):

Issue #155, opened 2018-04-26
Activity Register Button


catarinaamachado:
When enrolments are not allowed in activities, if you are not logged in the register button still shows up.
If you are logged in, it works perfectly and the button doesn't appear.

PR #285, opened 2023-07-03
Fix tests & update to ex_machina

danielsp45:
Some of the tests were failing duo to User fixtures being used in the tests, so I added ex-machina


mix.exs
```diff
@@ -33,31 +33,55 @@ defmodule Atomic.MixProject do
...
     [
-      {:bcrypt_elixir, "~> 3.0"},
       {:phoenix, "~> 1.6.14"},
-      {:phoenix_ecto, "~> 4.4"},
+
+      # core
+      {:esbuild, "~> 0.4", runtime: Mix.env() == :dev},
+      {:phoenix_live_reload, "~> 1.2", only: :dev},
+      {:phoenix_live_view, "~> 0.17.5"},
+      {:quantum, "~> 3.0"},
```
ruilopesm: Why is `quantum` needed?

danielsp45: I'm not sure, it was already there

mix.exs
```diff
@@ -33,31 +33,55 @@ defmodule Atomic.MixProject do
...
+
+      # database
       {:ecto_sql, "~> 3.6"},
+      {:phoenix_ecto, "~> 4.4"},
       {:postgrex, ">= 0.0.0"},
+
+      # security
+      {:bcrypt_elixir, "~> 3.0"},
+      {:argon2_elixir, "~> 3.0"},
```
ruilopesm: We should only use one of these algorithms. I think atomic is using bcrypt, but check it yourself.

mix.exs
```diff
@@ -33,31 +33,55 @@ defmodule Atomic.MixProject do
...
+
+      # Uploads
```
ruilopesm:
```suggestion
      # uploads
```

mix.exs
```diff
@@ -33,31 +33,55 @@ defmodule Atomic.MixProject do
   # Type `mix help deps` for examples and options.
   defp deps do
     [
-      {:bcrypt_elixir, "~> 3.0"},
       {:phoenix, "~> 1.6.14"},
-      {:phoenix_ecto, "~> 4.4"},
+
+      # core
+      {:esbuild, "~> 0.4", runtime: Mix.env() == :dev},
```
ruilopesm: This should be moved to `# frontend` category.

mix.exs
```diff
@@ -33,31 +33,55 @@ defmodule Atomic.MixProject do
...
-      {:flop, "~> 0.17.0"},
-      {:phoenix_live_dashboard, "~> 0.6"},
-      {:esbuild, "~> 0.4", runtime: Mix.env() == :dev},
+      {:swoosh, "~> 1.5"},
+      {:phoenix_swoosh, "~> 1.0"},
+
+      # frontend
       {:tailwind, "~> 0.1", runtime: Mix.env() == :dev},
+      {:flop, "~> 0.17.0"},
+      {:floki, ">= 0.30.0", only: :test},
```
ruilopesm: Why is this needed? `floki` specifically.

mix.exs
ruilopesm: Quantum is intended for cron jobs. We are planning to implement it in bokken ([#155](https://github.com/coderdojobraga/bokken/pull/155)), for example.
You can remove it for now 🙏🏼

ruilopesm: It is being used, my bad. Check this module: `Atomic.Quantum.CertificateDelivery`

MarioRodrigues10:
I think instead of doing a PR fixing user fixtures, you could fix all tests since you're already change a few of them, what you think?

ruilopesm:
@MarioRodrigues10 We should leave tests for new comers to make 👀
It is a very good way to understand how the platform works.

test/support/factories/accounts_factory.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  alias Atomic.Accounts.User
+
+  defmacro __using__(_opts) do
+    quote do
+      @roles ~w(admin staff student)a
+
+      def user_factory do
+        %User{
+          id: Faker.UUID.v4(),
```
MarioRodrigues10:
```suggestion
```

test/atomic_web/controllers/user_auth_test.exs
```diff
@@ -3,17 +3,16 @@ defmodule AtomicWeb.UserAuthTest do
...
-  import Atomic.AccountsFixtures
 
   @remember_me_cookie "_atomic_web_user_remember_me"
 
   setup %{conn: conn} do
     conn =
       conn
       |> Map.replace!(:secret_key_base, AtomicWeb.Endpoint.config(:secret_key_base))
-      |> init_test_session(%{})
+      |> Phoenix.ConnTest.init_test_session(%{})
```
MarioRodrigues10:
```suggestion
      |> init_test_session(%{})
```

test/atomic_web/controllers/page_controller_test.exs
```diff
@@ -1,8 +1,3 @@
 defmodule AtomicWeb.PageControllerTest do
   use AtomicWeb.ConnCase
-
-  test "GET /", %{conn: conn} do
-    conn = get(conn, "/")
-    assert html_response(conn, 200) =~ "Welcome to Phoenix!"
-  end
```
MarioRodrigues10:
```c
  test "GET /", %{conn: conn} do
    conn = get(conn, "/")
    assert html_response(conn, 200) =~ "Listing Activities"
  end
```

MarioRodrigues10 (changes requested):
Also there's some inconsistency because you erased some live tests and kept others.

MarioRodrigues10 (approved):

MarioRodrigues10:
Great work :rocket:

PR #288, opened 2023-07-12
Fix test action

MarioRodrigues10:


test/support/fixtures/activities_fixtures.ex
```diff
@@ -39,11 +39,10 @@ defmodule Atomic.ActivitiesFixtures do
   @doc """
   Generate a enrollment.
   """
-  def enrollment_fixture(attrs \\ %{}) do
-    {:ok, enrollment} =
-      attrs
-      |> Enum.into(%{})
-      |> Atomic.Activities.create_enrollment()
+  def enrollment_fixture(_attrs \\ %{}) do
```
danielsp45: I think we can delete this function, since we added the Event Factory with the Enrollments this function is no longer used in the code.

ruilopesm (approved):

PR #290, opened 2023-07-15
Remove unnecessary role

MarioRodrigues10:
After a meeting we had, we found that there's no need to have this role in users.


ruilopesm:
Explain the decision.

MarioRodrigues10:
Well there's only two roles needed, which are admin that is the person that overviews and manages the platform if needed, and a normal student that can create is own organization, or be part of  an organization.

We didn't found a reason where a staff is needed.

ruioliveira02:
I agree with @MarioRodrigues10, @RuiL1904. For me this can be merged right now

ruilopesm (approved):

PR #292, opened 2023-07-23
Protect Routes

MarioRodrigues10:
This PR belongs to @joaodiaslobo , in order to merge this quicker i fixed some tests and open it because @joaodiaslobo  will be out of town for a week and we can't delay this feature much more.


lib/atomic_web/live/activity_live/edit.ex
```diff
@@ -10,13 +10,21 @@ defmodule AtomicWeb.ActivityLive.Edit do
...
+        department.organization_id
+      end)
+
+    if organization_id in organizations do
+      {:noreply,
+       socket
+       |> assign(:page_title, gettext("Edit Activity"))
+       |> assign(:activity, activity)}
+    else
+      raise AtomicWeb.MismatchError
```
ruilopesm, line 25: Why are you raising an exception on an `handle_params` function?

lib/atomic/activities/speaker.ex
```diff
@@ -5,18 +5,23 @@ defmodule Atomic.Activities.Speaker do
...
     timestamps()
   end
 
   @doc false
   def changeset(speaker, attrs) do
     speaker
-    |> cast(attrs, [:name, :bio])
-    |> validate_required([:name, :bio])
+    |> cast(attrs, [:name, :bio, :organization_id])
+    |> validate_required([:name, :bio, :organization_id])
```
ruioliveira02: Move this to `@required_fields`

lib/atomic_web/live/activity_live/edit.ex
ruioliveira02: I think what he is doing is raising an exception making the current request fail with 404 status, which automatically shows the 404 page. Smart

ruioliveira02 (commented):
image

lib/atomic/activities.ex
```diff
@@ -22,6 +22,20 @@ defmodule Atomic.Activities do
...
+      |> apply_filters(opts)
+      |> Repo.all()
+      |> Repo.preload(:departments)
+
+    Enum.filter(activities, fn activity ->
+      Enum.any?(activity.departments, fn department ->
+        department.organization_id == organization_id
+      end)
+    end)
+  end
```
ruilopesm: This function would benefit from using Ecto queries. The final statement, of filtering, would be much faster if done within the database level.

lib/atomic_web/controllers/user_auth.ex
```diff
@@ -47,6 +48,14 @@ defmodule AtomicWeb.UserAuth do
     conn
   end
 
+  defp get_default_organization(user) do
+    if user do
+      List.first(Accounts.get_user_organizations(user))
+    else
+      nil
+    end
+  end
```
ruilopesm: Hmm, I don't like this. Mainly for two reasons: it is not guaranteed that a select onto the database will always return the results in the same order; I don't like we are not letting the user choose his default organization - I'm of the opinion that an user should have a `default_organization` field, capable of being edited through the aplicattions's UI.

lib/atomic_web/live/activity_live/edit.ex
```diff
@@ -10,13 +10,21 @@ defmodule AtomicWeb.ActivityLive.Edit do
...
-       :activity,
-       Activities.get_activity!(id, [:activity_sessions, :speakers, :departments])
-     )}
+  def handle_params(%{"organization_id" => organization_id, "id" => id} = _params, _url, socket) do
+    activity = Activities.get_activity!(id, [:activity_sessions, :speakers, :departments])
+
+    organizations =
+      Enum.map(activity.departments, fn department ->
+        department.organization_id
+      end)
```
ruilopesm: This can be made within the database level, too.

lib/atomic_web/live/activity_live/edit.ex
ruilopesm: Hmm, I see. Why not wrap this logic inside a function and call it on a Plug implemented at the route level? That way, we could also use that function when deciding to display (or not) the "Edit" button on the _show_ of an activity. Agreed? 🙏

Edit: This was supposed to be a reply to the already created thread 😅

ruioliveira02: I think that is a good idea. Will clean up the code quite a bit

lib/atomic_web/controllers/user_auth.ex
MarioRodrigues10: I agree with you, this require major changes in the code in order to do this

lib/atomic/activities.ex
```diff
@@ -22,6 +22,18 @@ defmodule Atomic.Activities do
...
+  def list_activities_by_organization_id(organization_id, opts) do
+    activities =
+      Activity
+      |> apply_filters(opts)
+      |> join(:inner, [a], d in assoc(a, :departments))
+      |> where([a, d], d.organization_id == ^organization_id)
+      |> select([a, _d], a)
+      |> Repo.all()
+
+    activities
```
ruioliveira02:
```suggestion
      Activity
      |> apply_filters(opts)
      |> join(:inner, [a], d in assoc(a, :departments))
      |> where([a, d], d.organization_id == ^organization_id)
      |> select([a, _d], a)
      |> Repo.all()
```

lib/atomic/activities.ex
```diff
@@ -41,6 +53,14 @@ defmodule Atomic.Activities do
     |> Repo.preload(preloads)
   end
 
+  def get_activity_organizations!(activity, _preloads \\ []) do
+    departments = Map.get(activity, :departments, [])
+
+    organization_ids = Enum.map(departments, & &1.organization_id)
+
+    organization_ids
```
ruioliveira02:
```suggestion
    Enum.map(departments, & &1.organization_id)
```

lib/atomic/organizations.ex
```diff
@@ -255,6 +262,24 @@ defmodule Atomic.Organizations do
...
+      [:member, :admin, :owner]
+
+  """
+  def roles_bigger_than_or_equal(role) do
+    case role do
+      :follower -> [:follower, :member, :admin, :owner]
+      :member -> [:member, :admin, :owner]
+      :admin -> [:admin, :owner]
+      :owner -> [:owner]
+    end
```
ruioliveira02:
```suggestion
    list = [:follower, :member, :admin, :owner]
    Enum.drop_while(list, fn elem -> elem != role end)
```

I am not too sure about this, but if we want to add more roles in the future, we can just insert in the list in order, and not need to do anything else to this function.

ruioliveira02 (commented):
image
image

There are two ways to go about this. We can fix this right now, or merge this and make it a problem for the person implementing that page

ruioliveira02 (approved):
:rocket:

PR #295, opened 2023-08-11
Fix verify function

MarioRodrigues10:


lib/atomic/organizations.ex
```diff
@@ -227,6 +227,28 @@ defmodule Atomic.Organizations do
...
+  ## Examples
+
+      iex> verify_admin(user, departments)
+      true
+
+      iex> verify_admin(user, departments)
+      false
+
+  """
+  def verify_admin(_user, []), do: false
```
ruilopesm:
```suggestion
  def is_admin?(_user, []), do: false
```
I prefer the old name 👀

ruilopesm (approved):

PR #304, opened 2023-08-11
refactor: improve coexistence between logged in and logged out users

ruilopesm:


MarioRodrigues10 (dismissed):
Good job :+1:

MarioRodrigues10:
Do it config.ex which handles which pages are in the sidebar should appear in here , or do you think that should be handled by who solves #311 ?

ruilopesm:
@MarioRodrigues10 I think #311 must be solved first, then.

ruilopesm:
Will be solved via #392.

PR #306, opened 2023-08-11
feat: add styling to activity certificates

ruioliveira02:
It is not perfect. There are no fonts configured, and there is a white border around the generated PDF. If someone could help me, that would be perfect. Otherwise I am hard stuck on this

~~I hate wkhtmltopdf~~


lib/atomic_web/views/helpers.ex
```diff
@@ -159,6 +159,15 @@ defmodule AtomicWeb.ViewUtils do
     |> URI.to_string()
   end
 
+  def display_enum(enum) do
```
MarioRodrigues10: Please make some doc tests for this function

lib/atomic/quantum/certificate_delivery.ex
```diff
@@ -30,37 +31,55 @@ defmodule Atomic.Quantum.CertificateDelivery do
...
+  defp process_enrollment(%Enrollment{} = enrollment) do
+    activity =
+      Repo.get_by!(Activity, id: enrollment.session.activity_id)
+      |> Repo.preload([:departments])
+
+    organizations =
+      activity.departments
+      |> Enum.map(fn d -> d.organization_id end)
+      |> Enum.dedup()
+      |> Enum.map(fn o -> Repo.get_by!(Organization, id: o).name end)
```
MarioRodrigues10: Departments are now associated to a session please remake this

lib/atomic/quantum/certificate_delivery.ex
```diff
@@ -30,37 +31,55 @@ defmodule Atomic.Quantum.CertificateDelivery do
...
-          require Logger
+    |> Enum.each(&process_enrollment/1)
+  end
 
-          Logger.error(
-            "Failed to generate certificate for enrollment #{enrollment.id} because: #{inspect(reason)}"
+  defp process_enrollment(%Enrollment{} = enrollment) do
+    activity =
+      Repo.get_by!(Activity, id: enrollment.session.activity_id)
+      |> Repo.preload([:departments])
```
MarioRodrigues10: There's a function that already does this no need to rewrite code.

MarioRodrigues10 (changes requested):

lib/atomic_web/templates/pdf/activity_certificate.html.eex
```diff
@@ -1,19 +1,73 @@
...
-        </div>
-    </body>
+<head>
+    <title>Certificado <%= @activity.title %></title>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width" />
+    <style>
+        /* Inline styles */
+        .container {
+            background: url('http://localhost:4000/images/atomic.png') no-repeat center;
```
MarioRodrigues10:
```suggestion
            background: url('http://localhost:4000/images/atomic.png') no-repeat center;
```
This won't work in prod

MarioRodrigues10 (dismissed):

lib/atomic_web/templates/pdf/activity_certificate.html.eex
ruioliveira02: It would because the rendering happens server side, and in the production server `localhost` is prod anyway. However I do agree it is not the best thing in the world, so I will change it when I have the time

lib/atomic/quantum/certificate_delivery.ex
```diff
@@ -30,37 +33,53 @@ defmodule Atomic.Quantum.CertificateDelivery do
...
+
+      {:error, reason} ->
+        # We log the error instead of just letting the function crash because we want the task to try and
+        # send the remaining certificates despite the failure. If the first certificate would
+        # fail to generate, we still want all the others to be sent
+        require Logger
+
+        Logger.error(
+          "Failed to generate certificate for enrollment #{enrollment.id} because: #{inspect(reason)}"
+        )
```
ruilopesm: In the future, we should think of a way to save this logs 💡

lib/atomic_web/views/helpers.ex
```diff
@@ -159,6 +159,34 @@ defmodule AtomicWeb.ViewUtils do
     |> URI.to_string()
   end
 
+  @doc ~S"""
```
ruilopesm, line 275:
```suggestion
  @doc """
```
> Elixir supports some sigil variants to deal with escaping characters and interpolation. In particular, uppercase letters sigils do not perform interpolation nor escaping. For example, although both ~s and ~S will return strings, the former allows escape codes and interpolation while the latter does not.

We don't need this there, then.

ruilopesm (changes requested):
Don't forget to change `@daily` to `0 0 * * *`, in order for the job to actually run at every midnight.

ruilopesm (approved):

PR #307, opened 2023-08-11
Update email confirmation

### LINKED ISSUE #303: Verify account with email after register
User can only log in into the plataform after the account is verified through email.

ruioliveira02:
Now the pages are styled and there is a plug which restricts access to routes to users that have already been confirmed via email

### DIFF
--- a/lib/atomic_web/router.ex
+++ b/lib/atomic_web/router.ex
@@ -44,7 +44,10 @@ defmodule AtomicWeb.Router do
   end
 
   scope "/", AtomicWeb do
-    pipe_through [:browser, :require_authenticated_user]
+    pipe_through [:browser, :require_authenticated_user, :require_confirmed_user]
+
+    get "/users/settings", UserSettingsController, :edit
+    put "/users/settings", UserSettingsController, :update
 
     live_session :logged_in, on_mount: [{AtomicWeb.Hooks, :authenticated_user_state}] do
       live "/", HomeLive.Index, :index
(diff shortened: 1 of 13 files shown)


lib/atomic_web/router.ex
```diff
@@ -155,9 +158,6 @@ defmodule AtomicWeb.Router do
 
   scope "/", AtomicWeb do
     pipe_through [:browser, :require_authenticated_user]
-
-    get "/users/settings", UserSettingsController, :edit
-    put "/users/settings", UserSettingsController, :update
     get "/users/settings/confirm_email/:token", UserSettingsController, :confirm_email
```
MarioRodrigues10, line 161: Is this route enough to confirm an account?
Shouldn't there be something like:
```
    get "/confirm", UserConfirmationController, :new
    post "/confirm", UserConfirmationController, :create
    get "/confirm/:token", UserConfirmationController, :edit
    post "/confirm/:token", UserConfirmationController, :update
```

MarioRodrigues10 (changes requested):
Also please make sure CI passes

MarioRodrigues10 (changes requested):
After i register this logs me in instead of redirecting to login page.
And also when im logged in i can only acess general routes.

The flow should be:
Register -> Confirm email -> Login

lib/atomic_web/router.ex
ruioliveira02: No

MarioRodrigues10 (approved):
:rocket:

PR #316, opened 2023-08-13
[WIP] fix: app flow

feliciofilipe:


feliciofilipe:
@MarioRodrigues10 can you finish this one plz?

MarioRodrigues10:
@feliciofilipe  I didn't finish this yet for 2 main reasons #310 wasn't approved (and i'm looking to finish that first) and because there's a problem related to this.
We use 2 types of states:
General user state -> for pages that everyone can see logged out or logged in
Authenticated user state -> for pages that you can only see when your logged in 

If we use general user state for some pages, we will have to verify for each of them if current_organization is present, which isn't the best approach. 
Also there's some details related to the sidebar

PR #321, opened 2023-08-17
feat: add user profile

### LINKED ISSUE #297: User profile page

joaodiaslobo:
Closes #297

### DIFF
--- a/lib/atomic/accounts/user.ex
+++ b/lib/atomic/accounts/user.ex
@@ -85,7 +85,10 @@ defmodule Atomic.Accounts.User do
     changeset
     |> validate_required([:handle])
     |> validate_format(:handle, ~r/^[a-zA-Z0-9_.]+$/,
-      message: "must only contain alphanumeric characters, numbers, underscores and periods"
+      message:
+        Gettext.gettext(
+          "must only contain alphanumeric characters, numbers, underscores and periods"
+        )
     )
     |> validate_length(:handle, min: 3, max: 30)
     |> unsafe_validate_unique(:handle, Atomic.Repo)
--- a/lib/atomic/accounts/user.ex
+++ b/lib/atomic/accounts/user.ex
@@ -84,7 +84,9 @@ defmodule Atomic.Accounts.User do
   defp validate_handle(changeset) do
     changeset
     |> validate_required([:handle])
-    |> validate_format(:handle, ~r/^[a-z0-9_]+$/, message: "must only contain lowercase characters, numbers, and underscores")
+    |> validate_format(:handle, ~r/^[a-z0-9_]+$/,
+      message: "must only contain lowercase characters, numbers, and underscores"
+    )
     |> validate_length(:handle, min: 3, max: 30)
     |> unsafe_validate_unique(:handle, Atomic.Repo)
     |> unique_constraint(:handle)
--- a/lib/atomic/accounts/user.ex
+++ b/lib/atomic/accounts/user.ex
@@ -84,8 +84,8 @@ defmodule Atomic.Accounts.User do
   defp validate_handle(changeset) do
     changeset
     |> validate_required([:handle])
-    |> validate_format(:handle, ~r/^[a-z0-9_]+$/,
-      message: "must only contain lowercase characters, numbers, and underscores"
+    |> validate_format(:handle, ~r/^[a-zA-Z0-9_.]+$/,
+      message: "must only contain alphanumeric characters, numbers, underscores and periods"
     )
     |> validate_length(:handle, min: 3, max: 30)
     |> unsafe_validate_unique(:handle, Atomic.Repo)
--- a/lib/atomic/accounts/user.ex
+++ b/lib/atomic/accounts/user.ex
@@ -136,6 +136,21 @@ defmodule Atomic.Accounts.User do
   end
 
   @doc """
+  A user changeset for changing the handle.
+
+  It requires the handle to change otherwise an error is added.
+  """
+  def handle_changeset(user, attrs) do
+    user
+    |> cast(attrs, [:handle])
+    |> validate_handle()
+    |> case do
+      %{changes: %{handle: _}} = changeset -> changeset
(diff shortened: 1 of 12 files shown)


priv/repo/migrations/20221014155230_create_users_auth_tables.exs
```diff
@@ -8,6 +8,7 @@ defmodule Atomic.Repo.Migrations.CreateUsersAuthTables do
       add :id, :binary_id, primary_key: true
       add :name, :string
       add :email, :citext, null: false
+      add :handle, :citext, null: false
```
MarioRodrigues10, line 11: handle should be unique, add that constraint

priv/repo/migrations/20221014155230_create_users_auth_tables.exs
```diff
@@ -20,7 +21,7 @@ defmodule Atomic.Repo.Migrations.CreateUsersAuthTables do
       timestamps()
     end
 
-    create unique_index(:users, [:email])
+    create unique_index(:users, [:email, :handle])
```
MarioRodrigues10: If you want both the "email" and "handle" columns to have unique values independently (i.e., not related to each other), you can create two separate unique indexes, each focusing on one column. This way, each column's values will be unique within itself, but they won't be tied to each other.

```c
create unique_index(:users, [:email])
create unique_index(:users, [:handle])
```

MarioRodrigues10:
Also make sure, to provide some unit tests to test this feature :bulb:

priv/repo/migrations/20221014155230_create_users_auth_tables.exs
joaodiaslobo: Thank you!

lib/atomic/accounts/user.ex
```diff
@@ -9,14 +9,15 @@ defmodule Atomic.Accounts.User do
   alias Atomic.Organizations.{Membership, Organization}
   alias Atomic.Uploaders.ProfilePicture
 
-  @required_fields ~w(email password)a
+  @required_fields ~w(email handle password)a
```
MarioRodrigues10, line 12: Altough handle is mandatory to have , it should be on optional_fields because when user register he doesn't have an handle.

test/atomic/accounts_test.exs
```diff
@@ -93,7 +93,7 @@ defmodule Atomic.AccountsTest do
   describe "change_user_registration/2" do
     test "returns a changeset" do
       assert %Ecto.Changeset{} = changeset = Accounts.change_user_registration(%User{})
-      assert changeset.required == [:password, :email]
+      assert changeset.required == [:password, :handle, :email]
```
MarioRodrigues10, line 115:
```suggestion
      assert changeset.required == [:password, :email]
```

lib/atomic/accounts/user.ex
```diff
@@ -79,6 +81,17 @@ defmodule Atomic.Accounts.User do
     |> unique_constraint(:email)
   end
 
+  defp validate_handle(changeset) do
+    changeset
+    |> validate_required([:handle])
+    |> validate_format(:handle, ~r/^[a-z0-9_]+$/,
+      message: "must only contain lowercase characters, numbers, and underscores"
```
MarioRodrigues10:
```suggestion
      message: "must only contain lowercase characters, numbers, and underscores"
```
It should also contain uppercase chars like twitter or github

lib/atomic/accounts/user.ex
joaodiaslobo: Wouldn't it be better if the user was prompted to choose a handle when they register? I can add that to the register form in this PR.

ruilopesm: @joaodiaslobo I think it makes all the sense.

lib/atomic/accounts/user.ex
```diff
@@ -79,6 +81,17 @@ defmodule Atomic.Accounts.User do
     |> unique_constraint(:email)
   end
 
+  defp validate_handle(changeset) do
+    changeset
+    |> validate_required([:handle])
+    |> validate_format(:handle, ~r/^[a-zA-Z0-9_.]+$/,
+      message: "must only contain alphanumeric characters, numbers, underscores and periods"
```
ruilopesm:
```suggestion
      message: gettext("must only contain alphanumeric characters, numbers, underscores and periods")
```
Maybe it would be `Gettext.gettext` 😅

lib/atomic/accounts/user.ex
```diff
@@ -79,6 +81,17 @@ defmodule Atomic.Accounts.User do
...
+    )
+    |> validate_length(:handle, min: 3, max: 30)
+    |> unsafe_validate_unique(:handle, Atomic.Repo)
```
ruilopesm, line 94:
```suggestion
    |> unsafe_validate_unique(:handle, Atomic.Repo, message: gettext("handle already in use")
```

ruilopesm: Nevermind, the function already throws a message error. Source: https://github.com/elixir-ecto/ecto/blob/1b8be19e155cf88caa4226b00a1027eeee494714/lib/ecto/changeset.ex#L2655

MarioRodrigues10:
This is looking good, what is missing to finish the PR?

joaodiaslobo:
> This is looking good, what is missing to finish the PR?

Just wanted to add some finishing touches, should be ready for review now!

ruilopesm (commented):
image
I don't like this edit button, but can't think of something else 😅
Do you have any idea?

joaodiaslobo:
> image I don't like this edit button, but can't think of something else 😅 Do you have any idea?

@RuiL1904  Using other buttons as reference we can also have something like this:

image

What do you think? 🤔

ruilopesm:
@joaodiaslobo Its way better. Push the code, in order to see it on mobile and review it 🙏🏼

ruilopesm (approved):
🚀

PR #329, opened 2023-08-19
feat: add news related live views

ruilopesm:


ruilopesm:
I think "News" is not a good name, for two main reasons:
1. The plural and singular are the same
2. It makes hard to read the code for the above reason

What do you guys (@cesium/atomic-23-24) think? Does anyone have a better name suggestion?

priv/repo/seeds/news.exs
```diff
@@ -1,14 +1,13 @@
...
 
   def run do
     seed_news()
   end
 
   def seed_news() do
-    case Repo.all(New) do
+    case Repo.all(News) do
       [] ->
         organizations = Repo.all(Organization)
```
danielsp45, line 13: Don't forget to change `%New{}` to `%News{}` in the for loop.
The seeds are failing.

danielsp45 (changes requested):

danielsp45:
image
The Edit and Delete buttons are looking a bit inconsistent in terms of positioning

ruilopesm:
@danielsp45 Can you check if the positioning is better now? 🙏🏼

danielsp45 (approved):
Great work 🚀

PR #333, opened 2023-08-19
fix: improve session number of entries & dates validation

danielsp45:


lib/atomic/activities/session.ex
```diff
@@ -48,6 +48,8 @@ defmodule Atomic.Activities.Session do
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
     |> validate_location()
+    |> validate_dates()
+    |> validate_entries_number()
```
ruilopesm, line 52: This check already exists. It's literally on the line below 😅

lib/atomic/activities/session.ex
```diff
@@ -59,6 +61,28 @@ defmodule Atomic.Activities.Session do
...
+    else
+      changeset
+    end
+  end
+
+  defp validate_dates(changeset) do
+    start = get_change(changeset, :start)
+    finish = get_change(changeset, :finish)
+
+    if start && finish && start > finish do
```
ruilopesm:
```suggestion
    if start && finish && Date.compare(start, finish) == :gt do
```

lib/atomic/activities/session.ex
danielsp45: I didn't notice but it wasn't working anyway. I still could create sessions ending before the starting date and with a maximum number of participants lower than the minimum

ruilopesm: Are you sure about the second one? Because there is a constraint at the database level - refer to https://github.com/cesium/atomic/blob/6e13421c81cf2cf3b8a939e917b9ced878618bef/priv/repo/migrations/20221023200102_create_sessions.exs#L22

ruilopesm: How have you tested it? On a forms? If so, maybe the error is not being handled

danielsp45: I tested on a forms. The way I did it now prevents it from happening and shows an error message to the user, better UX overall

ruilopesm: Nice 🤩

lib/atomic/activities/session.ex
```diff
@@ -59,6 +60,28 @@ defmodule Atomic.Activities.Session do
     |> cast_embed(:location, with: &Location.changeset/2)
   end
 
+  defp validate_entries_number(changeset) do
+    minimum_entries = get_change(changeset, :minimum_entries)
+    maximum_entries = get_change(changeset, :maximum_entries)
+
+    if minimum_entries && maximum_entries && minimum_entries > maximum_entries do
```
ruilopesm:
```suggestion
    if minimum_entries > maximum_entries do
```
If those values do not exist on the changeset, `get_change/2` will return `nil` (https://hexdocs.pm/ecto/Ecto.Changeset.html#get_change/3). And `nil > nil` is `false`. Also, the validation of required fields (such as `minimum_entries` and `maximum_entries`) is made before.
I think the code is prettier like so, too 👀

lib/atomic/activities/session.ex
```diff
@@ -59,6 +60,28 @@ defmodule Atomic.Activities.Session do
...
+    else
+      changeset
+    end
+  end
+
+  defp validate_dates(changeset) do
+    start = get_change(changeset, :start)
+    finish = get_change(changeset, :finish)
+
+    if start && finish && Date.compare(start, finish) == :gt do
```
ruilopesm, line 82:
```suggestion
    if Date.compare(start, finish) == :gt do
```

lib/atomic/activities/session.ex
```diff
@@ -59,6 +60,28 @@ defmodule Atomic.Activities.Session do
     |> cast_embed(:location, with: &Location.changeset/2)
   end
 
+  defp validate_entries_number(changeset) do
+    minimum_entries = get_change(changeset, :minimum_entries)
+    maximum_entries = get_change(changeset, :maximum_entries)
+
+    if minimum_entries > maximum_entries do
+      add_error(changeset, :maximum_entries, "must be greater than minimum entries")
```
ruilopesm:
```suggestion
      add_error(changeset, :maximum_entries, Gettext.gettext("must be greater than minimum entries"))
```

lib/atomic/activities/session.ex
```diff
@@ -59,6 +60,28 @@ defmodule Atomic.Activities.Session do
...
+      changeset
+    end
+  end
+
+  defp validate_dates(changeset) do
+    start = get_change(changeset, :start)
+    finish = get_change(changeset, :finish)
+
+    if start && finish && Date.compare(start, finish) == :gt do
+      add_error(changeset, :finish, "must be after starting date")
```
ruilopesm:
```suggestion
      add_error(changeset, :finish, Gettext.gettext("must be after starting date"))
```

ruilopesm (approved):
👏🏼

PR #337, opened 2023-08-20
feat: add cycle to courses schema

ruilopesm:


data/courses.txt
danielsp45, line 1: Is it supposed for this file to contain all of the courses?

ruilopesm: I guess so. Which one(s) is(are) missing?

danielsp45: From the most known ones, Medicina is missing.
And there must be missing a bunch of smaller ones, unless our scope is only UMinho degrees.

ruilopesm: Our scope, for now, is only UMinho degrees.

danielsp45: Then just add medicina and it must be all good

danielsp45 (dismissed):

data/courses.txt
```diff
@@ -1,241 +1,58 @@
...
-Programa doutoral em Ciências da Comunicação
-Programa doutoral em Design de Moda
-Programa doutoral em Economia
-Programa doutoral em Estudos Culturais
-Programa doutoral em Gestão
-Programa doutoral em História
-Programa doutoral em Marketing e Estratégia
-Programa doutoral em Modernidades Comparadas: Literaturas, Artes e Culturas
-Programa doutoral em Sociologia
+Ciências da Educação
```
danielsp45: Move this lower to the courses begginning with `C`

danielsp45 (approved):

PR #340, opened 2023-08-20
fix: bug where one must reload to see that is following an organization

ruilopesm:


lib/atomic_web/live/organization_live/show.ex
```diff
@@ -100,6 +95,14 @@ defmodule AtomicWeb.OrganizationLive.Show do
     end
   end
 
+  defp maybe_update_default_organization(user, organization) do
+    if user.default_organization_id == nil do
```
ruilopesm:
```suggestion
    if is_nil(user.default_organization_id) do
```

lib/atomic_web/live/organization_live/show.ex
```diff
@@ -60,24 +60,19 @@ defmodule AtomicWeb.OrganizationLive.Show do
...
       {:ok, _organization} ->
+        maybe_update_default_organization(current_user, socket.assigns.organization)
+
         {:noreply,
          socket
          |> put_flash(:success, "Started following " <> socket.assigns.organization.name)
-         |> push_redirect(to: Routes.organization_index_path(socket, :index))}
+         |> assign(:following, true)
+         |> push_patch(
+           to: Routes.organization_show_path(socket, :show, socket.assigns.organization.id)
```
danielsp45, line 70: Is this `push_patch` needed? isn't it redirecting to the same page the user is currently in?

danielsp45 (approved):

PR #360, opened 2023-08-23
feat: improve register flow

### LINKED ISSUE #311: Finish register form
The flow should be: register (with email and password) -> confirm email -> finish register details (such as name, default organization, etc).

joaodiaslobo:
Adds an account setup form to be filled after email confirmation, also improves the layout of some authentication pages. Closes #311


lib/atomic/accounts.ex
```diff
@@ -183,6 +201,24 @@ defmodule Atomic.Accounts do
...
+        ""
+
+  """
+  def extract_email_address_local_part(email) do
+    segments =
+      email
+      |> String.split("@")
+
+    List.first(segments)
+  end
```
ruilopesm: This should be inside an helpers module. Contexts are meant to work with the model part only.

ruilopesm (commented):
image
The way the error is being shown is weird 😅
Could you enhance it?

ruilopesm (commented):
Also, when an handle has already been taken, an exception is being raised, instead of letting the user know about it.

ruilopesm (approved):
What a great work! 🚀

PR #370, opened 2023-08-24
fix: enrollments update

ruilopesm:
Before: updating an enrollment was failing when the maximum number of enrollments had been reached.
Now: updating an enrollment (such as setting its presence to true) is working as intended.


MarioRodrigues10 (approved):
:+1:

PR #381, opened 2023-08-25
feat: add role admin when CeSIUM email register

### LINKED ISSUE #188: Default admin permissions for cesium's emails
A member of CeSIUM's direction suggested we could make any address that ends with `@cesium.di.uminho.pt` should be admin by default. 

We should probably restrict to department's emails: 

* comunicacao@cesium.di.uminho.pt
* drep@cesium.di.uminho.pt
* recreativo@cesium.di.uminho.pt
* caos@cesium.di.uminho.pt
* imagem@cesium.di.uminho.pt
* pedagogico@cesium.di.uminho.pt

And if we do implement super admin permissions as mention on issue #154 it should be given to:

* cesium@di.uminho.pt

MarioRodrigues10:


feliciofilipe:
I think it should not exist organizations accounts only users with temporary relations to organizations

MarioRodrigues10:
The purpose of this PR isn't for organization accounts, it is to give the role admin in the application for every cesium email

diogogmatos:
[Add admin role when registering a CeSIUM email](https://trello.com/c/RyKNjHrj/5-add-admin-role-when-registering-a-cesium-email)

lib/atomic_web/controllers/user_registration_controller.ex
```diff
@@ -16,7 +16,17 @@ defmodule AtomicWeb.UserRegistrationController do
 
   def create(conn, %{"user" => user_params}) do
     if user_params["password"] == user_params["password_confirmation"] do
-      case Accounts.register_user(user_params) do
+      [_username, domain] = String.split(user_params["email"], "@")
+
+      role =
+        case domain do
+          "cesium.di.uminho.pt" -> :admin
```
joaodiaslobo, line 23: Please, put this on an environment variable

ruilopesm:
I don't think we should implement the feature this way.

joaodiaslobo:
Outdated

PR #375, opened 2023-08-24
feat: add pagination and table components

joaodiaslobo:
Adds customizable table and pagination components. Uses [Flop Phoenix ](https://hexdocs.pm/flop_phoenix/0.14.2/Flop.Phoenix.html)for column filters inside the table. Fixes #367 , fixes #335 

image

PR #383, opened 2023-08-26
fix: list user sessions function

ruilopesm:


ruilopesm:
Will be solved by #375.

PR #388, opened 2023-08-26
fix: clean home page

feliciofilipe:


MarioRodrigues10 (dismissed):
This PR was approved because we decided to change home page to a feed

PR #402, opened 2023-09-01
feat: user settings page

### LINKED ISSUE #330: Improve user settings
I don't like it opening a new page and having a total different design. It would be better to follow something similar to what was made in [Parzival](https://github.com/joinum/parzival/tree/main/lib/parzival_web/live/app/profile_live).

### LINKED ISSUE #399: Change user handle to slug
See what @feliciofilipe said at #393 🙌🏼

joaodiaslobo:
Fixes #330 , fixes #399


ruilopesm (commented):
image
The way both errors are being shown is weird 😅

joaodiaslobo:
> image The way both errors are being shown is weird 😅

Would this be better @RuiL1904 ? 🤔

image

ruilopesm:
> > image The way both errors are being shown is weird 😅
> 
> Would this be better @RuiL1904 ? 🤔
> 
> image
> 

Yes!

lib/atomic_web/config.ex
```diff
@@ -99,7 +99,7 @@ defmodule AtomicWeb.Config do
         key: :profile,
         title: "Profile",
         icon: :user,
-        url: Routes.user_show_path(conn, :show, current_user.handle),
+        url: Routes.profile_show_path(conn, :show, current_user.slug),
```
ruilopesm: Since you are here, why not implement [this](https://hexdocs.pm/phoenix/Phoenix.Param.html). By setting `slug` as the `key`, we would only need to write `current_user` here 💡

lib/atomic_web/live/user_live/form_component.ex
```diff
@@ -5,7 +5,7 @@ defmodule AtomicWeb.UserLive.FormComponent do
 
   @impl true
   def update(%{user: user} = assigns, socket) do
-    changeset = Accounts.update_user(user)
+    changeset = Accounts.update_user(user, %{})
```
ruilopesm:
```suggestion
    changeset = Accounts.update_user(user)
```
Make `%{}` be the default value.

ruilopesm (commented):
A user should need to verify its email after changing it, by the way.

lib/atomic/accounts/user.ex
```diff
@@ -10,18 +10,21 @@ defmodule Atomic.Accounts.User do
...
   schema "users" do
     field :name, :string
     field :email, :string
-    field :handle, :string
+    field :slug, :string
     field :password, :string, virtual: true, redact: true
     field :hashed_password, :string, redact: true
     field :confirmed_at, :naive_datetime
 
+    field :phone_number, :string
```
MarioRodrigues10, line 27: Create a function which validates phone number

MarioRodrigues10 (dismissed):

lib/atomic/accounts/user.ex
ruilopesm: Check this issue (#403)

MarioRodrigues10 (approved):
:rocket:

PR #412, opened 2023-09-11
fix: problems with platform admin users

joaodiaslobo:
Some pages dependent on the `current_organization` key from socket assigns for redirects were causing multiple errors when being viewed by platform admin users, this PR changes these pages to use the organization related to the entity being viewed. Also makes sure that buttons to create new entities are only shown if an organization is actually selected.


lib/atomic_web/live/activity_live/edit.ex
```diff
@@ -11,14 +11,17 @@ defmodule AtomicWeb.ActivityLive.Edit do
 
   @impl true
   def handle_params(%{"id" => id}, _, socket) do
+    activity = Activities.get_activity!(id, [:speakers, :departments])
+    organization = activity.departments |> List.first() |> Map.get(:organization_id)
```
ruilopesm: What if the activity is organized by two or more different organizations?

joaodiaslobo: ~~I was not aware that this could be something that happened on the platform, I'll remove any kind of organization association from the activities in the routes and form component and make sure everything works fine without it. Thanks for the heads up!~~

It has been decided by the development team that for milestone 1.0.0 activities can only be organized by one organization. This PR will address that accordingly. 🚀

lib/atomic/activities/activity.ex
```diff
@@ -49,6 +50,7 @@ defmodule Atomic.Activities.Activity do
 
     has_many :enrollments, Enrollment, foreign_key: :activity_id
 
+    belongs_to :organization, Organization
```
MarioRodrigues10, line 49: An activity should belong to multiple organizations.
Also this association is already being done with ActivityDepartment

MarioRodrigues10 (dismissed):

lib/atomic/activities/activity.ex
ruilopesm: We already discussed this on our Slack.

ruilopesm (approved):

PR #415, opened 2023-09-13
feat: add timer to notifications

### LINKED ISSUE #404: Add timer for notifications
For example, after 5 seconds, a notification should be dismissed automatically.
In the future, we could have a settings page where the user could tailor this value to his needs 💡

joaodiaslobo:
Fixes #404 

notification


MarioRodrigues10 (approved):
:rocket:

PR #416, opened 2023-09-13
fix: security flaw with routes protection

### LINKED ISSUE #414: Security flaw with routes protection
The current plug protecting the routes is only checking if the authenticated user has permissions inside the organization in the `organization_id` parameter, while this does work on all occasions and redirects inside the app, if a malicious user decides to change the `organization_id` parameter to one belonging to an organization that they have priveleges over, the plug authorizes the user. In other words, the object's association with the id parameter is not being checked.

image

#292 Added `AtomicWeb.MismatchError` for this specific reason but currently it doesn't seem to be in use.

joaodiaslobo:
Fixes #414 

image


lib/atomic_web/plugs/verify_association.ex
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+      _ ->
+        conn
+    end
+  end
+
+  defp verify_association(entity, conn) do
+    if has_relation?(entity, conn.params["organization_id"]) do
+      conn
```
MarioRodrigues10, line 29:
```suggestion
      conn
      |> send_response(200, "Association verified")
      |> halt()
```

MarioRodrigues10 (dismissed):

lib/atomic_web/plugs/verify_association.ex
joaodiaslobo: This plug is used to authorize page access taking into account the parameters in the connection. Sending a 200 response wouldn't work here, because our goal is to keep the `conn` undisrupted in order to display the page.

image
_(this should show the partner page)_

MarioRodrigues10 (approved):

PR #417, opened 2023-09-14
refactor: enhance enrollments schema

### LINKED ISSUE #319: Fix enrollments confusion
Currently, there are two schemas called "enrollments" - one for an event and one for a session. There should only be one, that would work with both of the mentioned entities.

danielsp45:


MarioRodrigues10 (dismissed):
I think we should also change the function names, for example:
`list_enrollments`  to `list_activity_enrollments`  so everything stays under the same context

danielsp45:
> I think we should also change the function names, for example:
> 
> `list_enrollments`  to `list_activity_enrollments`  so everything stays under the same context

I think that's a bit redundant because you will do `Activities.list_enrollments()` and then `Events.list_enrollments()`. It will already be self explanatory which one I'm referring to.

ruilopesm (commented):
Everything looks perfect, but solve the remaining conflicts, please 🙏🏼

ruilopesm (approved):

PR #419, opened 2023-09-16
fix: update atomic logo

joaodiaslobo:
Removes all references to the old logo and replaces them with the new one, also makes sure stuff in the sidebar isn't user selectable.


MarioRodrigues10 (approved):
Good job :+1:

PR #426, opened 2023-09-22
build: upgrade phoenix and liveview

joaodiaslobo:
Update Phoenix to `1.7.0` and Phoenix Live View to `0.19.5`. Also replaces the icons package with another that only uses Heroicons.


lib/atomic_web.ex
```diff
@@ -102,7 +103,7 @@ defmodule AtomicWeb do
 
       alias Atomic.Uploaders
       alias AtomicWeb.Router.Helpers, as: Routes
-      alias Icons.Heroicons
+      alias Heroicons
```
nelsonmestevao: No need to alias something to itself 👀
```suggestion
```

joaodiaslobo: That's true 😅, thanks for the heads up!

lib/atomic_web/live/hooks.ex
```diff
@@ -3,6 +3,7 @@ defmodule AtomicWeb.Hooks do
   Ensures common `assigns` are applied to all LiveViews attaching this hook.
   """
   import Phoenix.LiveView
+  import Phoenix.Component
```
ruilopesm, line 6: Why is this import needed here?

joaodiaslobo: After Phoenix Live View 0.18.0 update, `assign/3` _(function used in this file)_ moved to `Phoenix.Component`, this move happened to multiple functions.

> You must import Phoenix.Component where you previously imported Phoenix.LiveView.Helpers when upgrading. You may also need to import Phoenix.Component where you also imported Phoenix.LiveView and some of its functions have been moved to Phoenix.Component.

[Read about it here](https://github.com/phoenixframework/phoenix_live_view/blob/main/CHANGELOG.md#0180-2022-09-20)

ruilopesm (commented):
As per the documentation says, we should use the `link/1` function, provided by `Phoenix.Component`.
image

ruilopesm (approved):

PR #440, opened 2023-10-03
refactor: add attributes to components

joaodiaslobo:
Uses the attribute and slot features introduced in `Phoenix LiveView 0.18.0` to clean up components code and make sure their usage is compile time verified.

### DIFF
--- a/lib/atomic_web/components/calendar.ex
+++ b/lib/atomic_web/components/calendar.ex
@@ -8,6 +8,13 @@ defmodule AtomicWeb.Components.Calendar do
   import AtomicWeb.Components.CalendarMonth
   import AtomicWeb.Components.CalendarWeek
 
+  attr :id, :string, default: "calendar", required: false
+  attr :current_path, :string, required: true
+  attr :activities, :list, required: true
+  attr :mode, :string, required: true
+  attr :timezone, :any, required: true
+  attr :params, :map, required: true
+
   def calendar(
         %{
           current_path: current_path,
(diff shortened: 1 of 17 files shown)


lib/atomic_web/components/calendar.ex
```diff
@@ -8,6 +8,13 @@ defmodule AtomicWeb.Components.Calendar do
   import AtomicWeb.Components.CalendarMonth
   import AtomicWeb.Components.CalendarWeek
 
+  attr :id, :string, default: "calendar", required: false
+  attr :current_path, :string, required: true
+  attr :activities, :list, required: true
+  attr :mode, :string, required: true
+  attr :timezone, :any, required: true
```
ruilopesm: Isn't there a type for this?

joaodiaslobo: The `attr` macro only supports the following atoms: `:boolean`, `:integer`, `:float`, `:string`, `:atom`, `:list`, `:map`, `:global` and structs. 
Since Timex doesn't have a defined struct for timezones I think the best idea here is to set the type to `:string`, since that's what is given to time functions (usually `"Europe/Lisbon"`)

lib/atomic_web/components/empty.ex
```diff
@@ -6,9 +6,13 @@ defmodule AtomicWeb.Components.Empty do
 
   alias Inflex
 
+  attr :id, :string, default: "empty-state", required: false
+  attr :placeholder, :string, required: true
+  attr :url, :atom, required: true
```
ruilopesm: Isn't this a string?

joaodiaslobo: yep, sorry!

ruilopesm (approved):

PR #445, opened 2023-11-10
feat: add trello integration action

diogogmatos:


.github/workflows/trello.yml
```diff
@@ -0,0 +1,82 @@
...
+          # Throw an error if no Trello cards can be found in the PR description.
+          # Default: false
+          github-require-trello-card: true
+
+          # Newline-separated list of mapping between Github username and Trello username.
+          # Use it for people who have different usernames in Github and Trello.
+          # If the current username is not in the list, we still try to find a Trello user with that username.
+          github-users-to-trello-users: |-
+            GithubUser1:TrelloUser1
+            GithubUser2:TrelloUser2
```
ruilopesm: Two things:
1. These should not be placeholders 😅
2. Creating this list will lead to a PR adding a GitHub username, Trello username mapping every time a collaborator joins CAOS, isn't there a better way?

diogogmatos: I didn't replace the placeholder cause I was waiting for your thoughts on the action before doing that boring work kk. And yeah, it's a bad way but I can't think of a way to do it better, there will always have to be some kind of map to associate github usernames with trello usernames, the only way would be to change the code behind the action itself so that it would find the trello user based on their name and not username.

diogogmatos: @ruilopesm the best solution would be everyone setting their username in trello to be exactly like the one in github, since the action always tries to find the trello user based on the github username.

diogogmatos:
[feat: add trello integration action](https://trello.com/c/i9csncQh/17-feat-add-trello-integration-action)

danielsp45 (approved):

PR #455, opened 2024-02-17
feat: add CVs

MarioRodrigues10:


lib/atomic/uploader.ex
```diff
@@ -5,7 +5,7 @@ defmodule Atomic.Uploader do
   """
 
   @versions [:original, :medium, :thumb]
-  @extensions_whitelist ~w(.svg .jpg .jpeg .png)
+  @extensions_whitelist ~w(.svg .jpg .jpeg .png .pdf .docx)
```
ruioliveira02, line 8: I don't think we should handle CV in this module. It defines functions to apply image transformations which do not make sense in a PDF or DOCX file.

priv/repo/migrations/20221014155230_create_users_auth_tables.exs
```diff
@@ -13,6 +13,7 @@ defmodule Atomic.Repo.Migrations.CreateUsersAuthTables do
       add :hashed_password, :string, null: false
       add :confirmed_at, :naive_datetime
       add :profile_picture, :string
+      add :cv, :string
```
ruioliveira02, line 16: Beware when modifying existing migrations. This is only OK because the project isn't in prod yet

ruioliveira02 (commented):
Change the test CV to something resembling a CV and not a project statement please :pray:

priv/repo/migrations/20221014155230_create_users_auth_tables.exs
MarioRodrigues10: I know that, but thanks for the advice !

MarioRodrigues10: I know that, but thanks for the advice !

lib/atomic/uploader.ex
MarioRodrigues10: Should i break this logic in this PR, or should i refactor the code in another?

ruioliveira02: This one

PR #457, opened 2024-02-20
feat: avatar component

joaodiaslobo:
Reusable avatar component for users and organizations


lib/atomic_web/components/avatar.ex
```diff
@@ -0,0 +1,49 @@
+defmodule AtomicWeb.Components.Avatar do
+  @moduledoc false
+  use AtomicWeb, :component
+
+  attr :src, :string, default: nil
+  attr :name, :string, required: true
+  attr :type, :atom, values: [:user, :company], default: :user
```
ruilopesm:
```suggestion
  attr :type, :atom, values: [:user, :company], default: :user
```
What about organizations?

joaodiaslobo: Organizations are companies, I chose to call it that name because its also has the same look as partner avatars

MarioRodrigues10 (changes requested):
Avatar at `organization_show_page` is rounded instead of squared.

ruilopesm (commented):
You forgot to replace avatars on `_live_navbar.html.heex` file.

MarioRodrigues10 (changes requested):
Please replace avatars at:

- Person at `organization_show_page`
- Speaker at  `activity_show_page`  and `activity_index_page`

MarioRodrigues10 (approved):
:rocket: :orange_heart:

PR #458, opened 2024-02-20
feat: add component for primary button

FilipeR13:


lib/atomic_web/components/button.ex
```diff
@@ -2,6 +2,16 @@ defmodule AtomicWeb.Components.Button do
   @moduledoc false
   use AtomicWeb, :component
 
+  attr :url, :string, required: true
+  slot :inner_block, required: true
```
ruilopesm, line 6:
```suggestion
  slot :inner_block, required: true

```

MarioRodrigues10 (changes requested):
Run `mix format` in order to pass CI.

MarioRodrigues10 (changes requested):
This looks awesome :star_struck: , there's only 2 buttons left:

- New Partner button at `partner_index_page`
- New Board member button at `board_index_page` 

Could you update them?
After that it's ready to go !!

MarioRodrigues10 (approved):
Good job !! :rocket: :orange_heart:

PR #464, opened 2024-02-22
refactor: page component

joaodiaslobo:
Makes it a functional component, improves documentation, ensures that the right border fills full height

### DIFF
--- a/lib/atomic_web/components/page.ex
+++ b/lib/atomic_web/components/page.ex
@@ -16,7 +16,7 @@ defmodule AtomicWeb.Components.Page do
   def page(assigns) do
     ~H"""
     <div class="flex min-h-full flex-col items-stretch justify-between lg:flex-row">
-      <div class={"#{if @bottom_border, do: 'border-b', else: ''} flex w-full flex-col bg-white lg:flex-row lg:border-r"}>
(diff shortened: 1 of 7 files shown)


lib/atomic_web/components/page.ex
```diff
@@ -1,24 +1,22 @@
...
   @moduledoc """
   Component for the main page layout.
-
-  * `title` - The title of the page.
-  * `bottom_border` - Whether to show a bottom border after the page header. Defaults to `false`.
-  * `actions` - Slot for actions to be rendered in the page header. An action could be, for example, a button to create a new record.
-  * `inner_block` - Slot for the body content of the page.
   """
-  use AtomicWeb, :live_component
+  use Phoenix.Component
```
ruilopesm, line 5:
```suggestion
  use AtomicWeb, :component
```

joaodiaslobo: This doesn't really work since the component is being imported for all pages.
See [this](https://elixirforum.com/t/trying-to-make-my-function-components-available-everywhere-getting-an-error-module-info-1-is-undefined/49680/15). Also written on the [docs](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html). In the future we should probably change all components to follow this.

ruilopesm (approved):

PR #467, opened 2024-02-22
fix: elixir version

ruilopesm:


MarioRodrigues10 (approved):
:rocket: :orange_heart:

PR #471, opened 2024-02-24
feat: add column enrolled to activities

### LINKED ISSUE #461: Add enrolled field to posts that are activities
### Check for existing issues

- [X] Completed

### Describe the bug / provide steps to reproduce it

Enrolled field is, somehow, absent on posts that are activities.

### What browsers are you seeing the problem on?

Firefox

### If applicable, add screenshots to help explain present your vision of the bug

image

FilipeR13:


ruilopesm (commented):
A few things:

- Max enrollments check on activity show page is not working properly, not showing the "Enroll" button even thought the activity has available space
- An _unenrollment_ its not taking effect on the `enrolled` field of its activity, it should decrease it, in this case

FilipeR13:
For performance reasons, I decided to keep the `:enrolled` atom on `show.ex` to increase the speed with which the page reloads and displays this same field. This is because, in this way, in the `reload()` method, the activity doesn't need to reload as a whole, but only the enrolled variable.

lib/atomic/activities/activity.ex
```diff
@@ -125,6 +127,37 @@ defmodule Atomic.Activities.Activity do
...
+    case {enrolled, maximum_entries} do
+      {nil, nil} ->
+        validate_enrollments_values(
+          changeset.data.enrolled,
+          changeset.data.maximum_entries,
+          changeset
+        )
+
+      {nil, maximum} ->
+        validate_enrollments_values(changeset.data.enrolled, maximum, changeset)
```
danielsp45, line 143: Why are you using `changeset.data.enrolled` instead of `enrolled` defined on the line 131?

lib/atomic/activities/activity.ex
```diff
@@ -125,6 +127,37 @@ defmodule Atomic.Activities.Activity do
...
+
+      {enrolled, nil} ->
+        validate_enrollments_values(enrolled, changeset.data.maximum_entries, changeset)
```
danielsp45, line 146: Same question as the previous comment

danielsp45 (dismissed):
Please merge with the `develop` branch 🙏🏻

lib/atomic/activities/activity.ex
FilipeR13: Here I followed the same logic that its used on function `validate_entries`  on line 74

lib/atomic/activities.ex
```diff
@@ -471,9 +492,8 @@ defmodule Atomic.Activities do
       0
   """
   def get_total_enrolled(activity_id) do
-    ActivityEnrollment
-    |> where(activity_id: ^activity_id)
-    |> Repo.aggregate(:count, :id)
+    activity = get_activity!(activity_id)
+    activity.enrolled
   end
```
ruilopesm: I think we don't need this function anymore.

lib/atomic/activities/activity.ex
ruilopesm: We should think on how to simplify both functions logic. For now, let's stick with this 🙏🏼

lib/atomic_web/live/activity_live/show.ex
```diff
@@ -82,6 +83,7 @@ defmodule AtomicWeb.ActivityLive.Show do
 
   defp reload(socket) do
     socket
+    |> assign(:enrolled, Activities.get_activity!(socket.assigns.id).enrolled)
```
ruilopesm: Instead of fetching the whole activity, add a named parameter, to `reload` function, like `action: :enroll` (or `action: :uneroll`) that would increment or decrement `enrolled` value accordingly. Then, change its call on `handle_info/2` 💡

ruilopesm (approved):
What a great work 🙌🏼

PR #479, opened 2024-03-06
feat: improve storybook

joaodiaslobo:
Changed it so that tailwind also checks classes used inside the storybooks.

Adds an icon list (icon atom names can be copied by clicking on the items):

image

New index page (removes all the default content):

image

Also made some adjustments to the sidebar tree


storybook/begin.story.exs
MarioRodrigues10, line 1: Fix image for mobile version
image

MarioRodrigues10 (changes requested):

storybook/begin.story.exs
joaodiaslobo: Damn! Nice catch, thank you! This could have been a major cause for concern!

MarioRodrigues10 (approved):
:rocket: :orange_heart:

PR #506, opened 2024-06-17
fix: log out functionality

Afonso-santos:
The log out functionality didn't work. Now I change the route  and it's work

### DIFF
--- a/lib/atomic_web/templates/layout/_live_navbar.html.heex
+++ b/lib/atomic_web/templates/layout/_live_navbar.html.heex
@@ -129,8 +129,8 @@
               <AtomicWeb.Components.Dropdown.dropdown
                 orientation={:top}
                 items={[
-                  %{name: "Your profile", link: Routes.profile_show_path(@socket, :show, @current_user)},
-                  %{name: "Sign out", link: Routes.user_session_path(@socket, :delete)}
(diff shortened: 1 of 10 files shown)


ruilopesm (dismissed):
This should actually be a `delete`. You need to change the link that is leading to this endpoint to have a `method="delete"` attribute 😄

lib/atomic_web/templates/layout/_live_navbar.html.heex
```diff
@@ -129,8 +129,8 @@
               <AtomicWeb.Components.Dropdown.dropdown
                 orientation={:top}
                 items={[
-                  %{name: "Your profile", link: Routes.profile_show_path(@socket, :show, @current_user)},
-                  %{name: "Sign out", link: Routes.user_session_path(@socket, :delete)}
+                  %{name: "Your profile", link: Routes.profile_show_path(@socket, :show, @current_user), method: "get"},
```
joaodiaslobo: You can now remove the `method` value from this map entry, since `"get"` is the default one.
```suggestion
                  %{name: "Your profile", link: Routes.profile_show_path(@socket, :show, @current_user)},
```

joaodiaslobo (changes requested):
Just remove the highlighted snippet and it should be good to go!

joaodiaslobo (approved):
Nice work, thanks!! 🚀

PR #513, opened 2024-08-05
feat: pre-commit hook

### LINKED ISSUE #342: Add pre-commit hook

MarioRodrigues10:
### Description

This pull request introduces a pre-commit hook to ensure code quality before commits are finalized. The hook performs the following checks:
 - Formatting
 - Linting
 - Testing

The setup for this pre-commit hook is handled by running `bin/setup`, which configures the hook whether you are working in a Docker environment or locally.

Additionally, this PR includes updates made to meet Credo CI requirements, which accounts for the additional changes beyond just the hook setup.


bin/pre-commit.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# pre-commit hook script
+
+set -e  
+
+echo 'Running pipeline...'
+
+mix checkCI
```
ruilopesm: Why not use the current `check` command?

MarioRodrigues10: `check` command has some bloat which will impede you to commit something if dependencies aren't ok, which is 
`deps.unlock --check-unused`.

bin/setup
```diff
@@ -74,6 +74,39 @@ else
...
+    local dir="$1"
+    while [ "$dir" != "/" ] && [ ! -d "$dir/.git" ]; do
+        dir="$(dirname "$dir")"
+    done
+    if [ -d "$dir/.git" ]; then
+        echo "$dir"
+    else
+        echo ""
+    fi
+}
```
ruilopesm: This can be replaced by simply running the command `git rev-parse --show-toplevel`

bin/pre-commit.sh
ruilopesm: You are right, didn't think about that 😅
Would you mind renaming the command to, for example, just `ci`? I kinda don't like the name being `checkCI`...

FilipeR13 (approved):

Issue #398, opened 2023-08-29
Refactor calendar component


ruilopesm:

joaodiaslobo:
Solved by #518

PR #518, opened 2024-08-08
refactor: improve calendar performance and style

nunom27:
#398


ruilopesm (commented):
Week view, in mobile, is not showing events, for some reason 🤔

nunom27:
> Week view, in mobile, is not showing events, for some reason 🤔

Do you actually follow the organization? You can try unfollowing and following the organization that has the activity again. I believe the seeds that makes generated users follow organizations is a bit broken, because the button shows up as following when the organization has 0 followers.

Also, does the same activity show up in any other view? For me it seems to be working.

ruilopesm:
@nunom27 I can see activities properly while on a desktop dimensions. After resizing the screen to mobile, activities do not show anymore.

I guess we'll have to wait until someone try to reproduce it 😅

joaodiaslobo:
Can't reproduce it on my side @ruilopesm . 🤔
On top of the mobile week view, make sure to select a day where there are events! (in this example, day 7 is already selected)

https://github.com/user-attachments/assets/7f3060c7-4476-4fad-a89c-6bb3316bc82f

MarioRodrigues10:
It works on my side, but looks a little bit odd on my side.
image
image

nunom27:
@MarioRodrigues10 That happened due to the activity starting at 23:35, so it only takes 5 rows until midnight (each row represents 5 minutes intervals). I can try, however, to put the title right in front of the time on those cases, so that it doesn't look as odd.

ruilopesm:
It's properly working for me now. I might have tested it wrongly 😅

lib/atomic_web/components/calendar/month.ex
```diff
@@ -35,55 +34,81 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
+                  </div>
                 </div>
-              </div>
-            </li>
-          </.link>
-        <% end %>
-      </ol>
-    </div>
+                <.link navigate={Routes.activity_index_path(AtomicWeb.Endpoint, :index)} class="bg-primary-600 flex-none self-center rounded-md px-3 py-2 font-semibold text-white hover:bg-primary-500 focus-visible:outline-primary-600 focus-visible:outline-2 focus-visible:outline-offset-2">
+                  Open activity
```
ruilopesm:
```suggestion
                  <%= gettext("Open activity") %>
```

lib/atomic_web/components/calendar/month.ex
```diff
@@ -35,55 +34,81 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
+                    </time>
+                  </div>
                 </div>
-              </div>
-            </li>
-          </.link>
-        <% end %>
-      </ol>
-    </div>
+                <.link navigate={Routes.activity_index_path(AtomicWeb.Endpoint, :index)} class="bg-primary-600 flex-none self-center rounded-md px-3 py-2 font-semibold text-white hover:bg-primary-500 focus-visible:outline-primary-600 focus-visible:outline-2 focus-visible:outline-offset-2">
```
ruilopesm: Why is this link going to the index page? Anyways, we already have a link covering the whole card, so I guess we wouldn't need a link here.

lib/atomic_web/components/calendar/month.ex
```diff
@@ -35,55 +34,81 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
+        Timex.shift(end_of_month, days: i)
+      end
+
+    if days_from_last_month == 1 do
+      current_month ++ next_month
+    else
+      past_month ++ current_month ++ next_month
+    end
+  end
+
```
ruilopesm: Could you document this function with a nice description and useful examples? I know it's private, but documentation would be nice for future contributors 👀

lib/atomic_web/components/calendar/month.ex
```diff
@@ -35,55 +34,81 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
 
     class =
       class_list([
-        {"relative py-2 px-3 lg:min-h-[110px] lg:flex hidden", true},
-        {col_start(weekday), index == 0},
-        {"bg-white", today? >= 0},
-        {"bg-zinc-50 text-zinc-500", today? < 0}
+        {"relative py-2 px-3 lg:h-28 lg:flex flex-col hidden", true},
+        {"bg-white", assigns.current_date.month == date.month},
+        {"bg-zinc-50 text-zinc-500", assigns.current_date.month != date.month}
```
ruilopesm: You can simply do this and ditch the use of that `class_list` function:
```html
<div class={["relative py-2 ...", assigns.current_date.month == date.month && "bg-white", ...]}>
    ....
</div>
```

lib/atomic_web/components/calendar/month.ex
```diff
@@ -99,38 +124,48 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
             </.link>
           </li>
         <% end %>
+        <%= if Enum.count(get_date_activities(@activities, @date)) > 2 do %>
+          <li class="text-zinc-500 hover:text-primary-600">
+            <button type="button" phx-click="show-more" phx-value-date={@date}>
+              +<%= Enum.count(get_date_activities(@activities, @date)) - 2 %> more
+            </button>
+          </li>
+        <% end %>
```
ruilopesm: There's a new short syntax for cases like this (same for other places, if you prefer it like so)
```
<li :if={Enum.count(get_date_activities(@activities, @date)) > 2} class="...">
```

<hr>

Also, would you consider this calculation heavy? If so, prefer to assign it before rendering the template, by calculating it only once (you are using it at least twice).

lib/atomic_web/components/calendar/month.ex
```diff
@@ -99,38 +124,48 @@ defmodule AtomicWeb.Components.CalendarMonth do
...
               "bg-zinc-900"
             end} text-white shirk-0"
           else
             if @today? == 0 do
               "text-orange-700"
             end
+            if @date.month != @current_date.month do
+              "text-zinc-500"
+            end
           end}"
```
ruilopesm: This is kinda confusing for a reader. Isn't there a better way to present it by using the syntax I've shown you before? 😄

lib/atomic_web/components/calendar/week.ex
```diff
@@ -116,9 +116,14 @@ defmodule AtomicWeb.Components.CalendarWeek do
...
-      <li class={"#{col_start(@idx + 1)} relative mt-px flex"} style={"grid-row: #{calc_row_start(activity.start)} / span #{calc_time(activity.start, activity.finish)}"}>
+      <% width = calc_total_overlaps(activity, @activities) + 1 %>
+      <% left = calc_left_amount(activity, @activities) %>
+      <li class={"#{col_start(@idx + 1)} relative mt-px flex"} style={"
+            grid-row: #{calc_row_start(activity.start)} / span #{calc_time(activity.start, activity.finish)};
+            width: #{(1/width)*100}%;
+            left: #{if left > 0 do (left / width) * 100 else 0 end}%"}>
         <.link patch={Routes.activity_show_path(AtomicWeb.Endpoint, :show, activity)}>
-          <div class="group absolute inset-1 flex flex-col overflow-y-auto rounded-lg bg-orange-50 p-2 text-xs leading-5 hover:bg-orange-100">
+          <div class={"#{if width != 1 do "sm:hover:z-10 sm:hover:w-max sm:max-w-[117px]" end} group absolute inset-1 flex flex-col overflow-x-hidden rounded-md bg-orange-50 p-2 text-xs leading-5 hover:bg-orange-100 sm:overflow-y-hidden sm:hover:overflow-y-auto"}>
```
ruilopesm: This is a known pitfall when using heex templates. Learn more about it [here](https://hexdocs.pm/phoenix_live_view/assigns-eex.html#pitfalls).

lib/atomic_web/components/calendar/week.ex
```diff
@@ -143,19 +148,58 @@ defmodule AtomicWeb.Components.CalendarWeek do
...
+    |> Enum.take_while(fn activity -> activity != current_activity end)
+    |> Enum.filter(fn activity ->
+      activity_interval = Timex.Interval.new(from: activity.start, until: activity.finish)
 
-    2 + 20 * time_diff
+      activity != current_activity && activity.start.day == current_activity.start.day and
+        Timex.Interval.overlaps?(current_interval, activity_interval)
+    end)
+    |> length()
   end
```
ruilopesm: Would you mind documenting these 3 functions too? 🥺

lib/atomic_web/live/calendar_live/show.html.heex
```diff
@@ -1,3 +1,84 @@
...
+        <div class="flex-1">
+          <div class="relative flex w-full items-center justify-between lg:flex-none">
+            <span class="text-base font-semibold text-zinc-900">
+              <%= if @mode == "month" do %>
+                <time datetime={@beginning_of_month}><%= Timex.format!(@beginning_of_month, "{Mshort} {YYYY}") %></time>
+              <% else %>
+                <%= case date_to_month(@beginning_of_week) == date_to_month(@end_of_week) do %>
+                  <% true -> %>
+                    <time datetime={@beginning_of_week}><%= Timex.format!(@beginning_of_week, "{Mshort} {YYYY}") %></time>
+                  <% _ -> %>
```
ruilopesm: Prefer to use an `if`, instead of a `case` here.

lib/atomic_web/views/calendar_utils.ex
```diff
@@ -5,24 +5,20 @@ defmodule AtomicWeb.CalendarUtils do
   use Phoenix.HTML
```
ruilopesm, line 5: Shouldn't this module be solely on the scope of the calendar? Or is it being used on other places?

lib/atomic_web/live/calendar_live/components/week.ex
```diff
@@ -0,0 +1,288 @@
...
+      "15H",
+      "16H",
+      "17H",
+      "18H",
+      "19H",
+      "20H",
+      "21H",
+      "22H",
+      "23H"
+    ]
```
joaodiaslobo:
```suggestion
    0..23 |> Enum.map(fn n -> "#{n}H" end)
```

joaodiaslobo (changes requested):

joaodiaslobo (approved):
Looks great!! 🎉📆

PR #520, opened 2024-08-19
feat: maps component

joaodiaslobo:
Google Maps component. Will be useful to display activity locations on their show page and might get used in other places in the future.
Made sure to create the respective storybook page (`/storybook/components/map`).

image


ruilopesm (approved):
🚀

PR #523, opened 2024-08-21
feat: enhance dropdown component with link navigation options and transitions

nunom27:


joaodiaslobo (changes requested):
Looks good! Very nice changes! 🎉 
If you can, please add module documentation to `lib/atomic_web/components/dropdown.ex` like it's done on the map component, thanks!

ruilopesm (dismissed):

joaodiaslobo (approved):
Awesome!

PR #526, opened 2024-08-22
refactor: sidebar

MarioRodrigues10:
## Description
The purpose of this PR is to refactor the sidebar, since it was using alpine and other outdated stuff.

## Related Issues
N/A

## Steps to reproduce or test
- mix phx.server
- Open at localhost:4000
- And see the sidebar in desktop and mobile views


lib/atomic_web.ex
```diff
@@ -33,6 +33,7 @@ defmodule AtomicWeb do
         root: "lib/atomic_web/templates",
         namespace: AtomicWeb
 
+      import AtomicWeb.Components.Sidebar
```
ruilopesm: This `import` will be injected into every view and that's not ideal. I would suggest you to reference sidebar's components by their full namespace. In the future, we can migrate our app to use the newest templates creation and then import this namespace there 🙌🏼

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,223 @@
+defmodule AtomicWeb.Components.Sidebar do
+  @moduledoc false
+  use AtomicWeb, :component
+
+  alias Phoenix.LiveView.JS
+  import AtomicWeb.Components.{Avatar, Dropdown, Icon}
+  alias Atomic.Organizations
+
+  attr :current_user, :map, required: true
+  attr :current_organization, :map, required: true
```
ruilopesm, line 10: Actually, these maps are structs, so we can type them using `User` and `Organization`, respectively, instead of just `:map` 😄

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,223 @@
...
+        </nav>
+      </div>
+    </div>
+    """
+  end
+
+  defp dropdown_items(current_user) do
+    [
+      %{
+        name: "Your profile",
```
ruilopesm:
```suggestion
        name: gettext("Your profile"),
```
Since you are on this 👀

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,223 @@
...
+        link: Routes.profile_show_path(AtomicWeb.Endpoint, :show, current_user)
+      },
+      %{name: "Sign out", link: Routes.user_session_path(AtomicWeb.Endpoint, :delete)}
```
ruilopesm:
```suggestion
      %{name: gettext("Sign out"), link: Routes.user_session_path(AtomicWeb.Endpoint, :delete)}
```

ruilopesm (commented):
Nice refactor. Just check my review and it's ready to merge.

Btw, I'm thinking that we need to find a better way to organize our components under the `components` folder. It's not a big deal, but it's starting to be a bit messy 😅

lib/atomic_web/components/sidebar.ex
MarioRodrigues10: Unfortunately non of those can be typed since sometimes their value is nil

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,205 @@
...
+  defp show_mobile_sidebar(js \\ %JS{}) do
+    js
+    |> JS.show(
+      to: "#mobile-sidebar",
+      transition:
+        {"ease-in duration-1000", "transform translate-x-0", "transform -translate-x-full"}
+    )
+    |> JS.show(to: "#sidebar-overlay")
+    |> JS.dispatch("focus", to: "#mobile-sidebar")
+  end
```
joaodiaslobo: The mobile sidebar animation looks very wrong. 🤔

https://github.com/user-attachments/assets/44f1e21b-9603-4c77-8bb0-af9451e723e2

joaodiaslobo (changes requested):
Why is the user dropdown not aligned to the bottom anymore? Is this intended?

image

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,205 @@
...
+      }
+    ]
+  end
+
+  defp show_mobile_sidebar(js \\ %JS{}) do
+    js
+    |> JS.show(
+      to: "#mobile-sidebar",
+      transition:
+        {"ease-in duration-1000", "transform translate-x-0", "transform -translate-x-full"}
```
joaodiaslobo:
```suggestion
        {"transition ease-in-out duration-300 transform", "-translate-x-full", "translate-x-0"}
```

joaodiaslobo (changes requested):
Also, make sure that when the user clicks on the dark area on mobile, the sidebar closes.

lib/atomic_web/components/sidebar.ex
joaodiaslobo: This fixes the problem with the closing animation

MarioRodrigues10:
Thanks @joaodiaslobo  🙌 

I removed a lot of bloat from div classes and somehow i moved the user up in the sidebar, I'll get that fixed today

lib/atomic_web/components/sidebar.ex
```diff
@@ -0,0 +1,209 @@
...
+                "#{if @current_page == page.key do
+                  "text-orange-500"
+                else
+                  "text-zinc-400 group-hover:text-orange-500"
+                end} h-6 w-6 shrink-0"
+              } />
+            <%= page.title %>
+          <% end %>
+        </li>
+      <% end %>
```
joaodiaslobo: Can this be a `<.link>` ?

joaodiaslobo (approved):
Nice refactor, thanks!

PR #530, opened 2024-09-11
refactor: improve announcements pages

### LINKED ISSUE #435: Update announcements page (user & admin) frontend

JoaoCoelho2003:
- [x] Fixed the image uploader component;
- [x] Refactor the existing announcement pages;
- [x] Made the announcements page exclusive to members of a publishing organization;
- [x] Improve the announcement feed design.

## New Announcement Page

The design of the _New Announcement_ page has been improved, and in the process, the previously malfunctioning image uploader was fixed.

image

## Announcements Page Access

Previously, the announcements page was a general feed available to all users. This functionality has been updated:

- The page is now exclusive to organization members, allowing them to manage announcements within their respective organizations;
- All routes leading to the announcements page have been revised. To access these pages, routes must now follow the format: `/organizations/{organization_id}/announcements`

[Screencast from 2024-11-21 00-52-13.webm](https://github.com/user-attachments/assets/078a4c71-c32c-4605-af1b-0f2e132ad995)

## Announcements Feed Page Changes

The tabs at the top of the announcements feed page were removed since the page is no longer used as a public feed but as a monitoring tool.

image

#435


ruilopesm (commented):
There's a lack of padding on every page. Make sure the padding obeys the one _imposed_ by the `page` component. In the future, we should make page's body be padded automatically 💡

FilipeR13:
I think it would be cool if the show page followed the same design as the edit/new announcement page when there’s an image to display. What do you think?

MarioRodrigues10:
We can have 2 approaches for this scenario:
image

We either limit the characters at index page and if the user wants the full information should open it and see the whole announcement context, or we can keep the whole text and remove the show page since won't be necessary anymore.


Which one do you like more @JoaoCoelho2003 ?

I would also like your input for this case @joaodiaslobo @ruilopesm . :raised_hands:

JoaoCoelho2003:
I personally prefer limiting the number of lines shown per announcement on the respective pages @MarioRodrigues10.

joaodiaslobo:
@MarioRodrigues10 
> I would also like your input for this case @joaodiaslobo @ruilopesm . 🙌

Keeping the whole text in the feed is definitely not the way to go but I also don't think we should limit what the organizations can write way too much (there still needs to exist a limit, but a bigger one). 
Maybe we should truncate the text on the post and allow the users to see the full content in the show page by clicking on the it.

MarioRodrigues10:
@joaodiaslobo There's already a lot of text being truncated at the application in different ways, in order to achieve that I think @JoaoCoelho2003 can create a general function at `helpers.ex` and truncate for this case only.

In the future, other people should reuse that function and truncate their text aswell if they need.

JoaoCoelho2003:
@joaodiaslobo 

> Keeping the whole text in the feed is definitely not the way to go but I also don't think we should limit what the organizations can write way too much (there still needs to exist a limit, but a bigger one). Maybe we should truncate the text on the post and allow the users to see the full content in the show page by clicking on the it.

In my opinion, we should truncate the text in the feed and, as Mario said, for the user to see the full post he would have to press it and check the announcement. About the text size limit, probably put a limit of around 500 characters or something like that.

image

The text size limit varies a lot depending on the platform, but since this is an announcement, and announcements typically don’t require much text as they are meant to be brief, we could consider setting the limit lower.

MarioRodrigues10:
I think 500 characters could be too much, check Instagram UI/UX for posts in either desktop or mobile I really like what they did, they truncate based on the text size.

ruilopesm:
We already have a function for truncating text, if I'm not mistaken. `maybe_slice_string/2` (we could change its name to `truncate` now that I think of it) 🙏

lib/atomic_web/live/announcement_live/components/announcement_card.ex
```diff
@@ -0,0 +1,41 @@
+defmodule AtomicWeb.AnnouncementLive.Components.AnnouncementCard do
+  @moduledoc false
+
+  import AtomicWeb.Components.Avatar
+
+  use AtomicWeb, :component
+
+  def announcement_card(assigns) do
```
FilipeR13, line 8: image
Should be a bottom border in the last announcement

lib/atomic_web/live/announcement_live/components/announcement_card.ex
```diff
@@ -0,0 +1,41 @@
...
+          <div>
+            <p class="text-sm font-medium text-zinc-700"><%= @announcement.organization.name %></p>
+            <p class="text-xs text-gray-500">
+              <span class="sr-only">Published on</span>
+              <time><%= relative_datetime(@announcement.inserted_at) %></time>
+            </p>
+          </div>
+        </div>
+        <div class="px-4 py-2">
+          <p class="text-lg font-semibold text-zinc-900" title={@announcement.title}>
```
FilipeR13, line 25: The activitys component is using `text-gray` instead of `text-zinc`. This is a issue in all code base, we should normalize this in another PR

lib/atomic_web/live/announcement_live/components/announcement_card.ex
```diff
@@ -0,0 +1,41 @@
...
+            <%= @announcement.title %>
+          </p>
+          <p class="mt-2 text-sm leading-relaxed text-zinc-500">
+            <%= @announcement.description %>
+          </p>
+        </div>
+        <%= if @announcement.image do %>
+          <div class="h-auto w-full overflow-hidden">
+            <img class="h-full w-full rounded-xl object-cover" src={Uploaders.Post.url({@announcement.image, @announcement}, :original)} alt="Announcement Image" />
```
FilipeR13: image
I think in the mobile version the image doenst look really good with this level of rounded

lib/atomic_web/live/announcement_live/index.html.heex
```diff
@@ -1,47 +1,24 @@
...
-      </.tab>
-    </.link>
-
-    <.link patch="?tab=following" replace={false}>
-      <.tab active={@current_tab == "following"}>
-        <%= gettext("Following") %>
-      </.tab>
-    </.link>
-  </.tabs>
+  <.tabs></.tabs>
```
FilipeR13: This solution to add the border in the page its not that good 😅, but I dont really know other solution. @joaodiaslobo any idea?

FilipeR13 (changes requested):

lib/atomic_web/live/announcement_live/index.html.heex
joaodiaslobo: The `page` component actually has a parameter to do this, called `bottom_border`. Unfortunately, it appears this functionality stopped working on a previous PR 😅.
@JoaoCoelho2003  Feel free to fix it and use it on this PR by making the following change in the `page` component:
```heex
def page(assigns) do
    ~H"""
    <div class="flex min-h-full flex-col items-stretch justify-between lg:flex-row">
      <div class="min-h-[100vh] flex w-full flex-col bg-white lg:flex-row lg:border-r">
        <main class="relative z-0 mb-10 flex-1 overflow-y-auto focus:outline-none xl:order-last">
          <div class={["mx-auto max-w-5xl px-4 sm:px-6 lg:px-8", @bottom_border && "border-b"]}>
            <div class="my-6 flex min-w-0 flex-row items-center justify-between"}>
              <h1 class="flex-1 select-none truncate text-2xl font-bold text-gray-900">
                <%= @title %>
              </h1>
              <%= render_slot(@actions) %>
            </div>
          </div>

          <%= render_slot(@inner_block) %>
        </main>
      </div>
    </div>
    """
  end
```

And use it like this:
```
<.page title="Announcements" bottom_border>
...
```

By the way, this PR is looking real good! 🤩 
Good job! 💪

lib/atomic_web/live/announcement_live/show.html.heex
```diff
@@ -1,22 +1,42 @@
...
-          <%= @announcement.title %>
-        </h4>
+<.page title="Announcements" bottom_border={true}>
+  <:actions>
+    <%= if @has_permissions? do %>
+      <.button navigate={Routes.announcement_edit_path(@socket, :edit, @announcement.organization, @announcement)} icon="hero-pencil-solid"><%= gettext("Edit Announcement") %></.button>
+    <% end %>
+  </:actions>
+  <div id={@announcement.id} class="mx-auto max-w-5xl">
+    <.link navigate={Routes.announcement_show_path(@socket, :show, @announcement)}>
```
FilipeR13: You shouldn't have a link to its own announcement on the show page.

lib/atomic_web/live/announcement_live/show.html.heex
```diff
@@ -1,22 +1,42 @@
...
-  <div class="px-4 flex flex-col gap-2 mt-6 sm:flex-row sm:items-center">
-    <div>
-      <div class="flex justify-between items-center">
-        <h4 class="text-2xl font-bold text-zinc-800">
-          <%= @announcement.title %>
-        </h4>
+<.page title="Announcements" bottom_border={true}>
+  <:actions>
+    <%= if @has_permissions? do %>
+      <.button navigate={Routes.announcement_edit_path(@socket, :edit, @announcement.organization, @announcement)} icon="hero-pencil-solid"><%= gettext("Edit Announcement") %></.button>
```
FilipeR13: imagem
As you can see, I can edit the announcement even though I am in a different organization.

FilipeR13: imagem
The button should dissapear when seing as user and not as organization. If I click, this error is raised, is probably because of that
imagem

FilipeR13: Replace the call to Routes.announcement_edit_path with the appropriate `~p` string.

lib/atomic_web/live/announcement_live/show.html.heex
```diff
@@ -1,22 +1,42 @@
...
+            <span class="py-4 text-zinc-800">
+              <%= Enum.map(String.split(@announcement.description, "\n"), fn phrase -> %>
+                <%= phrase %>
+              <% end) %>
+            </span>
+          </p>
+        </div>
+        <%= if @announcement.image do %>
+          <div class="h-auto w-full overflow-hidden mt-4">
+            <img class="h-full w-full object-cover md:rounded-xl" src={Uploaders.Post.url({@announcement.image, @announcement}, :original)} alt="Announcement Image" />
```
FilipeR13: You need to limit the height of the image or this might happen:
imagem
Try following the same strategy used on the main page

FilipeR13 (changes requested):

lib/atomic_web/live/announcement_live/show.html.heex
```diff
@@ -1,22 +1,40 @@
-<div class="py-4 px-4 sm:px-0">
-  <div class="px-4 flex flex-col gap-2 mt-6 sm:flex-row sm:items-center">
-    <div>
-      <div class="flex justify-between items-center">
-        <h4 class="text-2xl font-bold text-zinc-800">
+<.page title="Announcements" bottom_border={true}>
+  <:actions>
+    <%= if @has_permissions? do %>
+      <.button navigate={~p"/organizations/#{@announcement.organization}/announcements/#{@announcement}/edit"} icon="hero-pencil-solid"><%= gettext("Edit Announcement") %></.button>
```
FilipeR13, line 4: There are two previous issues that still need to be fixed:
The button should disappear when viewing as a user or another organization.

View as a user and not organization:
imagem

View as other organization:
imagem

FilipeR13 (changes requested):

lib/atomic_web/live/announcement_live/show.html.heex
JoaoCoelho2003: This should be fixed, unless you are using a master account

JoaoCoelho2003: image

Exactly, supposedly it is already working (let's not comment about the image used as an example)

FilipeR13: You are correct in that sense.

However, when viewing as a master in user mode, the button appears—which may be fine—but attempting to edit raises this error:
imagem

The first strategy to address this should be fixing the error. If that's not possible, consider hiding the button.

lib/atomic_web/live/announcement_live/show.html.heex
```diff
@@ -1,22 +1,40 @@
...
+    <% end %>
+  </:actions>
+  <div id={@announcement.id} class="mx-auto max-w-5xl">
+    <div class="mt-4 md:mt-2 flex flex-col justify-center rounded-lg bg-white lg:p-4">
+      <div class="flex items-center space-x-2 px-4 py-1">
+        <div class="flex-shrink-0">
+          <.avatar name={@announcement.organization.name} color={:light_zinc} size={:md} type={:organization} src={Uploaders.Logo.url({@announcement.organization.logo, @announcement.organization}, :original)} />
```
FilipeR13, line 11: There should be a link to the organization

FilipeR13 (changes requested):

FilipeR13 (approved):

PR #535, opened 2024-09-23
chore: upgrade to verified routes

ruilopesm:
## Description
Properly upgraded the whole application to make use of verified routes. This mechanism was introduced on Phoenix version 1.7 (the one we are currently using) and you can learn more about it and its advantages [here](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html).

Be careful while reviewing this PR, btw 😄


joaodiaslobo (approved):
LGTM!

Issue #539, opened 2024-10-28
Blocking the access to the camera in scanner freezes the application


FilipeR13:
If the user refuses to give access to the camera, the application freezes and the user can't leave the page of scanner.

![image.png](https://github.com/user-attachments/assets/e4c4cfff-f93e-4c35-affa-f9ecf3e6605f)

Fix the error and ensure the user doesn't get stuck.

ruilopesm:
This could help: https://github.com/cesium/seium.org/pull/729

PR #548, opened 2025-02-10
fix: scanner breaking if permission to use the camera is denied/blocked

JoaoCoelho2003:
#539 

If the permission to use the camera was denied in the scanner page then the user would not be able to go to any other page of the application (only after a f5 was it possible) because an error occurred due to a missing check in the scanner's logic.

image


JoaoCoelho2003:
The problem seems to be fixed now, as I thought it was just a simple missing check in the scanner's logic

FilipeR13 (approved):

